PR #5, opened 2022-12-01
Fix year and month function to truncate time component properly

shulhi:
Previously, the `year` function didn't truncate the time component. For example, running the function given a time equivalent to `December 1, 2022 1:18:10 AM` returns `January 1, 2022 1:18:10 AM`.

```ocaml
open Time in year (toTime (seconds 1669857490))
```

It should actually return `January 1, 2022 12:00:00 AM`.

Same deal with `month` function.


goodlyrottenapple (approved):
LGTM

PR #8, opened 2022-12-22
Update Nix instructions

ngua:
Adds some more details to the readme regarding setting up IOG's binary cache

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,10 @@ Accepting these prompts will set the required configuration values for you. Mark
 
 If you would prefer to configure the binary caches manually, you can do so by following the instructions [here](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started.html#setting-up-the-binary-cache).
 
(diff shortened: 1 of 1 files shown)


albertov (approved):

README.md
```diff
@@ -50,9 +50,34 @@ do you want to permanently mark this value as trusted (y/N)? y
 
 ```
 
-If you see the prompts above, terminate the command and first add IOG's binary caches to your Nix configuration as described [here](https://input-output-hk.github.io/haskell.nix/tutorials/getting-started.html#setting-up-the-binary-cache).
+Accepting these prompts will set the required configuration values for you. Marking them as trusted will ensure that they are used for future `nix` invocations in this repository.
```
siddharth-krishna, line 53: @ngua accepting the prompts did not work for me. I had the same issue described here: https://github.com/NixOS/nix/issues/6672

Was my problem the thing to do with trusted users below?

ngua: Possibly, if you aren't a trusted user

PR #9, opened 2022-12-23
Improve unification error when it fails

shulhi:
- Previously, this returns an empty list when no two types match. 
- Also, this handles returning error with dummy location when carried type errors are empty. Otherwise, when the carried type errors are empty, this function just returns an empty list instead of unification error.

### DIFF
--- a/inferno-core/src/Inferno/Infer.hs
+++ b/inferno-core/src/Inferno/Infer.hs
@@ -43,7 +43,7 @@ import Data.Bifunctor (bimap)
 import qualified Data.Bimap as Bimap
 import Data.Either (partitionEithers, rights)
 import Data.Generics.Product (HasType, getTyped, setTyped)
-import Data.List (find, unzip4) -- intercalate
+import qualified Data.List as List
 import qualified Data.List.NonEmpty as NEList
 import qualified Data.Map as Map
 import qualified Data.Map.Merge.Lazy as Map
@@ -95,7 +95,7 @@ import Inferno.Types.Type
     Substitutable (..),
     TCScheme (..),
     TV,
-    TypeClass (TypeClass),
+    TypeClass (..),
     typeBool,
     typeDouble,
     typeInt,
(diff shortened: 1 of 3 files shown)


siddharth-krishna (approved):

inferno-core/src/Inferno/Infer.hs
```diff
@@ -381,7 +381,7 @@ inferTypeReps allTypeClasses (ForallTC tvs tyCls (ImplType _impl ty)) inputTys o
         Left errs -> Left errs
         Right subst ->
           let tyClsSubst = Set.map (apply subst) tyCls
-           in case find (\(TypeClass nm _) -> nm == "rep") $ Set.toList tyClsSubst of
+           in case List.find (\(TypeClass nm _) -> nm == "rep") $ Set.toList tyClsSubst of
```
goodlyrottenapple, line 384: hmm not sure why i wasn't just using set find here?? probably not terribly important tho

inferno-core/src/Inferno/Infer.hs
```diff
@@ -1158,10 +1158,16 @@ unifies err (TSeries t1) (TSeries t2) = unifies err t1 t2
 unifies err (TOptional t1) (TOptional t2) = unifies err t1 t2
 unifies err (TTuple ts1) (TTuple ts2)
   | length (tListToList ts1) == length (tListToList ts2) = unifyMany err (tListToList ts1) (tListToList ts2)
+  | List.null err =
+      let pos = initialPos ""
+          loc = (pos, pos)
+       in throwError [UnificationFail Set.empty (TTuple ts1) (TTuple ts2) loc]
   | otherwise = throwError [UnificationFail (getTypeClassFromErrs err) (TTuple ts1) (TTuple ts2) loc | loc <- (getLocFromErrs err)]
```
goodlyrottenapple, line 1165: Looking at this code, the error throwing doesn't really make sense. My original design is definitely wonky here. The idea behind throwing a list was to link several places in code which generate a contradiction. e.g. somewhere we assign a variable the value bool and elsewhere we try to add a number to it. the error should point to those two places in the code as a contradiction. However, I'm not sure this ever properly worked as intended. It seems we always just throw a singleton error anyway.

To get proper error location, we should probably track location for types we are unifying and then add those to the error when throwing, instead of adding a dummy `pos`. This will require a bit of a re-engineering tho.

goodlyrottenapple (changes requested):
Think we should overhaul how position information is passed around to the unification function to not rely on dummy positions

PR #11, opened 2022-12-29
Fix broken vs code extension build + lsp server

goodlyrottenapple:
The nix build of the VS code inferno LSP extension was broken.
This also re-introduces the LSP server binary and re-routes the logging back to stderr, since the server communicates via stdio with the VS Code extension. 
Finally, this also includes some M1 specific tweaks due to some broken packages:
 - ormolu is broken on the M1, failing with a segfault when linking the executable
 - `ghc884` is not supported on the M1, the earliest supported one being `ghc8107`

### DIFF
--- a/inferno-lsp/inferno-lsp.cabal
+++ b/inferno-lsp/inferno-lsp.cabal
@@ -28,7 +28,8 @@ library
       src
   ghc-options: -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
   build-depends:
-      base                               >= 4.13 && < 4.17
+      aeson
+    , base                               >= 4.13 && < 4.17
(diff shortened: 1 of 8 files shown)


smurphy8 (approved):

inferno-lsp/inferno-lsp.cabal
```diff
@@ -70,4 +71,6 @@ executable inferno-lsp-server
   ghc-options: -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates -threaded -rtsopts -with-rtsopts=-T
   build-depends:
       base >=4.7 && <5
+    , inferno-core
```
siddharth-krishna, line 74: Btw, we need version bounds on all dependencies in order to upload to hackage. (We can add this before the next update)

PR #20, opened 2023-01-09
Profiling build of `inferno` binary

goodlyrottenapple:
This adds a profiling build of the inferno binary, which can be used via

```
nix shell .#inferno-core:exe:inferno-ghc924-prof
```

However, I've had to update `haskell.nix`, as the profiling build needs `recursion-schemes` >=5.2.2.3, due to:

> Workaround for https://gitlab.haskell.org/ghc/ghc/-/issues/18320, which was preventing code calling makeBaseFunctor from being profiled. 

The update to `haskell.nix` seems to have broken the dev shell??

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -88,12 +88,16 @@
               compiler
               # TODO
               # Do we want to enable any `crossPlatforms` here?
-              ((infernoFor compiler pkgs { }).flake { })
+              ((infernoFor { inherit compiler pkgs; }).flake { })
             ) ++
           lib.lists.forEach [ defaultCompiler ]
             (compiler: lib.attrsets.nameValuePair
               (compiler + "-prof")
-              ((infernoFor compiler pkgs { profiling = true; ghcOptions = [ "-eventlog" ]; }).flake { })
+              ((infernoFor {
+                inherit compiler pkgs;
+                profiling = true;
+                ghcOptions = [ "-eventlog" ];
+              }).flake { })
             )
         );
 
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -118,10 +123,16 @@
         };
         modules = [
           {
+            enableLibraryProfiling = profiling;
             packages = {
               # This takes forever to build
               ghc.components.library.doHaddock = false;
             };
+            packages.inferno-core = {
```
ngua, line 131: Could you format with `nixpkgs-fmt` for consistency with the rest of the Nix code please?

ngua (dismissed):
I can try to have a look at the broken haskell.nix breakages. Maybe bumping the `index-state` will be sufficient?

flake.nix
```diff
@@ -88,15 +88,20 @@
...
               # TODO
               # Do we want to enable any `crossPlatforms` here?
-              ((infernoFor compiler pkgs).flake { })
+              ((infernoFor compiler pkgs {}).flake { })
+            ) ++
+          lib.lists.forEach [ defaultCompiler ]
+            (compiler: lib.attrsets.nameValuePair
+              (compiler + "-prof")
+              ((infernoFor compiler pkgs {profiling = true; ghcOptions = [ "-eventlog" ];}).flake { })
             )
```
ngua, line 97: Do we only want profiling enabled on GHC 9.x? If that's the case we could drop the `forEach` and `nameValuePair` and merge the attribute sets directly

goodlyrottenapple: there are some new GHC9 only profiling modes, e.g. info-tables stuff, so depends on what profiling we want to do. I was just being lazy here tho, because the ghc8x version doens't work on M1 anyway

flake.nix
```diff
@@ -88,15 +88,20 @@
...
+              ((infernoFor compiler pkgs { profiling = true; ghcOptions = [ "-eventlog" ]; }).flake { })
             )
         );
 
       # This should be parameterized by the `pkgs` used to build the project. We
       # want users who get packages from our `overlays.default` to be able to
       # use their own `nixpkgs` without having to instantiate ours as well
       # (which would happen if we just use `self.packages` directly in the overlay)
-      infernoFor = compiler: pkgs: pkgs.haskell-nix.cabalProject {
+      infernoFor = compiler: pkgs: { ghcOptions ? [ ], profiling ? false }: pkgs.haskell-nix.cabalProject {
```
ngua, line 104: I think it would be best to make all of the arguments part of the attribute set now 
```suggestion
      infernoFor = { compiler, pkgs, ghcOptions ? [ ], profiling ? false }: pkgs.haskell-nix.cabalProject {
```

flake.nix
ngua: In that case I think it's OK to only have the profiling build for GHC 9, so we can remove the other `forEach`

PR #24, opened 2023-01-11
Add `inferno` cache and switch CI to build with Nix

### LINKED ISSUE #21: Set up binary cache and use Nix on CI
If we set up a binary cache (hopefully using Cachix), the following tasks also need to be done:
- [ ] Rewrite the GH Actions config to use Nix and Cachix actions and build everything in the flake
- [ ] Update the docs to instruct users on how to use Cachix
- [ ] Possibly set the `substituters` and `trusted-public-keys` for the Cachix cache in the flake directly so users don't need to install the Cachix CLI

ngua:
Closes #21 

There are some instructions in the readme on how to enable the new Inferno cache for local development.

CI will now build with Nix and push the artifacts to the cache. If you configure the cache locally you can pull them, making most Nix rebuilds much faster.

CI jobs will generally be faster. They'll only be slower (albeit considerably slower) when large updates happen with dependencies (e.g. a new haskell.nix revision)

I also disabled HLS in the dev shells (this is unrelated to the changes in this PR but I discovered it while making the changes here; see #25 for more details)


ngua:
Going to add the `devShells` to `packages.default` as well. I've found a few issues with the `devShells` today and it would be nice if these problems were caught on CI

siddharth-krishna (approved):
Not a nix expert, but looks good on a high level. Super excited for the cache!

Issue #10, opened 2022-12-27
Ormolu segfaults on M1 mac


goodlyrottenapple:
Trying to enter the dev shell via `nix develop` on an M1 Mac segfaults with a clang linker segfault building `ormolu`:

```bash
➜  inferno git:(main) nix develop
warning: Using saved setting for 'extra-substituters = https://cache.iog.io' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=' from ~/.local/share/nix/trusted-settings.json.
error: builder for '/nix/store/a00q1phmpqq1bd9d9l256lafjqgks69f-ormolu-exe-ormolu-0.5.0.1.drv' failed with exit code 1;
       last 10 log lines:
       > Configuring executable 'ormolu' for ormolu-0.5.0.1..
       > building
       > Preprocessing executable 'ormolu' for ormolu-0.5.0.1..
       > Building executable 'ormolu' for ormolu-0.5.0.1..
       > [1 of 2] Compiling Paths_ormolu     ( dist/build/ormolu/autogen/Paths_ormolu.hs, dist/build/ormolu/ormolu-tmp/Paths_ormolu.o, dist/build/ormolu/ormolu-tmp/Paths_ormolu.dyn_o )
       > [2 of 2] Compiling Main             ( app/Main.hs, dist/build/ormolu/ormolu-tmp/Main.o, dist/build/ormolu/ormolu-tmp/Main.dyn_o )
       > Linking dist/build/ormolu/ormolu ...
       > /nix/store/48py6zrawzim9ghrnkqwm36jl4j1l23x-clang-wrapper-11.1.0/bin/ld: line 256: 56158 Segmentation fault: 11  /nix/store/5wvlj00dr22ivh210b18ccv1i60h6c1q-cctools-binutils-darwin-949.0.1/bin/ld ${extraBefore+"${extraBefore[@]}"} ${params+"${params[@]}"} ${extraAfter+"${extraAfter[@]}"}
       > clang-11: error: linker command failed with exit code 139 (use -v to see invocation)
       > `clang' failed in phase `Linker'. (Exit code: 139)
       For full logs, run 'nix log /nix/store/a00q1phmpqq1bd9d9l256lafjqgks69f-ormolu-exe-ormolu-0.5.0.1.drv'.
[...]

PR #26, opened 2023-01-12
Improve formatting setup and fix `devShells.ghc884`

### LINKED ISSUE #22: Switch to `treefmt-nix`
Currently we're using `treefmt` to handle formatting. The developers recently split most of the Nix stuff into a separate repo with a `flakeModule` output that automates some things we're doing manually now. Also, the current `formatter` output isn't ideal in that it requires several executables to be on the `PATH` when called (`treefmt-nix` fixes that)

### LINKED ISSUE #23: `devShells.ghc884` is broken
I haven't had a chance to look into this very thoroughly, but I believe that `devShells.ghc884` is broken:
- `ormolu-0.5.0.1` requires `base >=4.14 && <5.0`
- `base-4.13` was bundled with GHC 8.8.4 
- `base` is a non-upgradeable package

Since `ormolu` is obtained via `haskell-nix.tool`, we get an incompatible build plan.

Possible solutions:
- drop `devShells.ghc884` altogether (not ideal)
- use an older `ormolu` version (also not ideal)
- don't include `ormolu` in `devShells` with older GHC versions
- switch to `haskellPackages.ormolu_0_5_0_1` from `nixpkgs` (maybe the best solution?)

ngua:
Closes #22, closes #23 and improves the necessary workarounds for #10 (previously Ormolu would have been built by `checks.formatting` and `packages.default` even on `aarch64-darwin`, now it's never built)


albertov (approved):
Looks good

PR #31, opened 2023-02-07
Define an Arbitrary instance for VCObject for golden tests

siddharth-krishna:
The `inferno-vc-server` executable serializes `VCObject`s into files, and we don't have golden tests for these. This PR adds the appropriate `Arbitrary` type class instances needed to generate golden files.

I've moved the `Arbitrary Expr` instance from `test/Parse/Spec.hs` into a new module `Inferno.Utils.Arbitrary`. This is because it needs to be used both by the tests in `inferno-core` as well as the `Arbitrary VCObject` instance in `inferno-vc`. Unfortunately, this is for some reason a dependency conflict... not sure why yet.

To do:
- [x] Fix the cabal dependency conflict
- [x] Fix the TODOs in the code


inferno-core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Revision History for inferno-core
 
+## 0.1.2 -- 2023-02-09
```
mirko-plowtech, line 3: @siddharth-krishna can we start using this convention? [semantic versioning](https://semver.org/)

siddharth-krishna: Sure!

inferno-core/src/Inferno/Instances/Arbitrary.hs
```diff
@@ -0,0 +1,629 @@
...
+              <*> ( do
+                      k <- choose (0, n)
+                      sequence [(,Nothing) <$> arbitrarySized (n `div` 3) | _ <- [1 .. k]]
+                  )
+              <*> arbitrary,
+            arbitraryArrayComp n,
+            One <$> arbitrary <*> arbitrarySized (n `div` 3),
+            Empty <$> arbitrary,
+            arbitraryAssert n,
+            -- arbitraryCase n -- Broken!
```
mirko-plowtech, line 516: **FIXME: test: "parseExpr and pretty are inverse up to normalizeExpr" hangs if we use them!**

inferno-core/src/Inferno/Instances/Arbitrary.hs
```diff
@@ -0,0 +1,629 @@
...
+                          xs
+                    )
+                  $ IntMap.toList baseOpsTable
+              )
+          <*> (arbitrarySized $ n `div` 3)
+          <*> arbitrary
+          <*> arbitrary
+      
+      -- FIXME: Currently broken, if we include this one we hang parseExpr test
+      arbitraryCase n =
```
mirko-plowtech, line 450: **here!**

inferno-core/src/Inferno/Instances/Arbitrary.hs
mirko-plowtech: done!

PR #38, opened 2023-02-28
Add profiling instructions to readme

siddharth-krishna:


README.md
```diff
@@ -196,6 +196,22 @@ To run a formatting check that will fail if any files are not formatted, run `ni
 
 **NOTE**: Ormolu currently segfaults during compilation on `aarch64-darwin` (M1 Macs). `ormolu` is accordingly omitted from the formatter (`formatters.aarch64-darwin`) and formatting check (`checks.aarch64-darwin.treefmt`). CI will still fail on unformatted Haskell sources as it runs on `x86_64-linux`, so it is recommended to install `ormolu-0.5.0.1` on your system using an alternate source. See https://github.com/plow-technologies/inferno/issues/10 for more.
 
+### Profiling
+
+We have a profiling build of the inferno binary, which can be used via:
+
+```
+nix shell .#inferno-core:exe:inferno-ghc924-prof
```
siddharth-krishna, line 204: Hmm I just tried this, and it built a lot of stuff but then didn't run anything. And trying to run `./result` after resulted in `Permission denied`. Should this command be `nix build ...`?

ngua: `nix shell` is used to temporarily add an executable to your `PATH`. For example, `nix shell .#inferno-lsp:exe:inferno-lsp-server` would add `inferno-lsp-server` to your path for the remainder of your current shell session. It doesn't add a `result` symlink

ngua: So in this case, since the name of the executable is `inferno`, it should add `inferno` to your `PATH`

ngua: It should probably be `nix run` if you want to build and run at the same time

README.md
```diff
@@ -196,6 +196,22 @@ To run a formatting check that will fail if any files are not formatted, run `ni
...
+```
+
```
ngua, line 206:
````suggestion

```
$ nix run .#inferno-core:exe:inferno-ghc924-prof 
```

Or equivalently:

```
$ nix build .#inferno-core:exe:inferno-ghc924-prof
$ ./result/bin/inferno
```

````

PR #42, opened 2023-04-14
[inferno-core] Added `toBCD` and `fromBCD` to Prelude

LTibbetts:
Example script:
```inferno
let d = Option.reduce id 0.0 (latestValue input0)
in let oBCD = fromBCD (toWord64 (round d))
in match oBCD with {
| None -> toWord64 0
| Some x -> x
}
```

### DIFF
--- a/inferno-core/src/Inferno/Module/Prelude.hs
+++ b/inferno-core/src/Inferno/Module/Prelude.hs
@@ -589,15 +589,11 @@ module Word
   @doc Convert a word to an `int`;
   fromWord : forall 'a. {requires fromWord on 'a} => 'a -> int := ###fromWordFun###;
 
-  define fromBCD on word64;
-
   @doc Decode a BCD-encoded word64;
-  fromBCD : forall 'a. {requires fromBCD on 'a} => 'a -> option of word64 := ###fromBCDFun###;
-
-  define toBCD on word64;
+  fromBCD : word64 -> option of word64 := ###fromBCDFun###;
 
   @doc Encode a BCD-encoded word64;
-  toBCD : forall 'a. {requires toBCD on 'a} => 'a -> word64 := ###toBCDFun###;
+  toBCD : word64 -> word64 := ###toBCDFun###;
 
 module Base
 


inferno-core/src/Inferno/Module/Prelude.hs
```diff
@@ -587,6 +589,16 @@ module Word
   @doc Convert a word to an `int`;
   fromWord : forall 'a. {requires fromWord on 'a} => 'a -> int := ###fromWordFun###;
 
+  define fromBCD on word64;
+
+  @doc Decode a BCD-encoded word64;
+  fromBCD : forall 'a. {requires fromBCD on 'a} => 'a -> option of word64 := ###fromBCDFun###;
```
siddharth-krishna, line 595: You don't need the type class `fromBCD` unless you're going to allow the function to be applied to more than the one type `word64`, right?

inferno-core/src/Inferno/Module/Prelude/Defs.hs
```diff
@@ -503,3 +507,33 @@ textSplitAt =
...
+    digitSeparation acc _ x | x == 0 = Just acc
+    digitSeparation acc offset x =
+      let d = x .&. 0b1111
+          acc' = acc + d * offset
+       in if d <= 9
+            then acc' `seq` digitSeparation acc' (offset * 10) (unsafeShiftR x 4)
+            else Nothing
+
+fromBCDFun :: Word64 -> Maybe Word64
+fromBCDFun = fmap fromIntegral . fromBCD . fromIntegral
```
siddharth-krishna, line 539: If you wanted to test these implementations, you could add some tests to `inferno-core/test/Eval/Spec.hs`

siddharth-krishna (commented):
Some minor suggestions

inferno-core/src/Inferno/Module/Prelude/Defs.hs
Daniel-Diaz: @siddharth-krishna They already have their own tests in plow-math: https://github.com/plow-technologies/all/blob/master/plow-math/test/Main.hs#L55

That's where this code comes from.

siddharth-krishna: Ah I see, cool! I suppose ideally we'd use the plow-math implementation here too, but then we'd have to open source plow-math too, would that make sense? (cc @smurphy8 )

Daniel-Diaz: I wouldn't open source the whole of plow-math, but I don't mind splitting the BCD stuff out and making that public.

smurphy8: We could make a dev side issue to do that but I don’t want to mess with it right now.

PR #48, opened 2023-06-19
LSP: error if script evaluates to a function

siddharth-krishna:
I added a check in the LSP that checks if the script type returns a function (by checking if it expects more inputs than input parameters supplied). This should warn users and make it less likely that they confuse the input parameters that are selected in the frontend UI with the parameters of a lambda function.

### DIFF
--- a/inferno-lsp/src/Inferno/LSP/ParseInfer.hs
+++ b/inferno-lsp/src/Inferno/LSP/ParseInfer.hs
@@ -25,7 +25,7 @@ import Inferno.Module.Prelude (ModuleMap, baseOpsTable, builtinModulesOpsTable, 
 import Inferno.Parse (InfernoParsingError, parseExpr, parseType)
 import Inferno.Parse.Commented (insertCommentsIntoExpr)
 import Inferno.Parse.Error (prettyError)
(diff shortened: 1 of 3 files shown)


inferno-lsp/src/Inferno/LSP/ParseInfer.hs
```diff
@@ -508,77 +508,55 @@ parseAndInfer prelude idents txt validateInput = do
...
+    checkScriptIsNotAFunction signature parameters =
+      -- A function with N parameters should have a signature a_1 -> a_2 -> ... -> a_{N+1}
+      if length signature > (length parameters + 1)
+        then Left [errorDiagnosticInfer 0 0 0 2 $ renderDoc $ vsep ["This script evaluates to a function. Did you mean to add input parameters instead?"]]
+        else Right ()
 
-    isTArr :: InfernoType -> Bool
-    isTArr = \case
-      TArr _ _ -> True
-      _ -> False
```
siddharth-krishna, line 575: @LTibbetts I simplified these functions by using the inbuilt function `collectArrs` that converts an `InfernoType` potentially containing `TArr`s to a `[InfernoType]`. I couldn't think of a reason why input type validation should be ignored if the final output is an array, so I removed that check. (In fact, don't we disallow array outputs in OnPing anyway?)

Issue #51, opened 2023-07-11
Fix slow parsing of tuple patterns and types


siddharth-krishna:
Example script:
```ocaml
fun x -> match x with {
  | ((((((((((((x)))))))))))) -> 2
}
```

<img width="346" alt="image" src="https://github.com/plow-technologies/inferno/assets/10712637/d86fce88-daee-4bec-ad5b-e8a2ea0b0dc9">

We should use `tupleElems` instead of `tupleArgs` (slower) in `Parse.hs` for tuple patterns and type lists as well.

PR #52, opened 2023-07-11
Add a timeout to LSP's parseAndInfer

siddharth-krishna:
This PR adds a timeout to the LSP so that if parsing or type checking hangs/takes too long, the editor doesn't break too. A timeout results in an error diagnostic, so that if the user modifies the script, the LSP can process the new version:

image

(This example takes too long to parse, see #51 )

### DIFF
--- a/inferno-lsp/src/Inferno/LSP/Server.hs
+++ b/inferno-lsp/src/Inferno/LSP/Server.hs
@@ -31,7 +31,7 @@ import qualified Data.Text.Utf16.Rope as Rope
 import Data.Time.Clock (UTCTime, getCurrentTime)
 import qualified Data.UUID.V4 as UUID.V4
 import Inferno.LSP.Completion (completionQueryAt, filterModuleNameCompletionItems, findInPrelude, identifierCompletionItems, mkCompletionItem, rwsCompletionItems)
-import Inferno.LSP.ParseInfer (parseAndInfer)
+import Inferno.LSP.ParseInfer (errorDiagnostic, parseAndInfer)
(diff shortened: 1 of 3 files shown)


inferno-lsp/src/Inferno/LSP/Server.hs
```diff
@@ -212,6 +204,29 @@ sendDiagnostics :: J.NormalizedUri -> J.TextDocumentVersion -> [J.Diagnostic] ->
...
+  (InfernoType -> Either T.Text ()) ->
+  m2 ParsedResult
+parseAndInferWithTimeout beforeParse afterParse prelude idents doc_txt validateInput = do
+  ts <- liftIO getCurrentTime
+  uuid <- UUID <$> liftIO UUID.V4.nextRandom
+
+  liftIO $ beforeParse (uuid, ts)
+  result <- do
+    -- Timeout parsing and type checking after 10 seconds
+    mResult <- liftIO $ timeout (10 * 1000000) $ parseAndInfer @m1 @_ @c prelude idents doc_txt validateInput
```
albertov, line 224: I recently learnt about `NumericUnderscores` which is great for readability of these big constants, lets you write ten million as `10_000_000`.

albertov (approved):
Looks good, just left a minor FYI comment

inferno-lsp/src/Inferno/LSP/Server.hs
siddharth-krishna: Ah, that's very nice, thanks :)

PR #55, opened 2023-07-11
Add array indexing and pattern matching

siddharth-krishna:
This PR adds support for array indexing `[1, 2, 3] !! 2 == Array.get [1, 2, 3] 2 == 3` and array pattern matching:
```
fun a -> match a with { | [] -> 0 | [x] -> 1 | [_, y] -> 2 | _ -> 3 }
```

Currently we only allow literal patterns like `[x, _, ...]`, but no cons-style patterns like Haskell's `x : xs`. This means you must have a final `| _ -> ...` or `| a -> ...` pattern in order for the pattern match to be exhaustive.


inferno-core/src/Inferno/Infer.hs
```diff
@@ -1031,17 +1034,50 @@ infer expr =
...
+                      (vars2, csPs) <- aux t ps
+                      let inferredTy = TArray t
+                      attachTypeToPosition
+                        patLoc
+                        TypeMetadata
+                          { identExpr = patternToExpr $ bimap (const ()) (const ()) pattern,
+                            ty = (Set.empty, ImplType Map.empty $ inferredTy),
+                            docs = Nothing
+                          }
+                      return (inferredTy, vars1 ++ vars2, csP `Set.union` csPs)
```
siddharth-krishna, line 1068: Type checking for array patterns is similar to that of tuple patterns, except every element of the array must have the same type (constraint `tIst'` below), so I modified `mkPatConstraint` to also return a set of type constraints.

inferno-core/src/Inferno/Infer.hs
```diff
@@ -1031,17 +1034,50 @@ infer expr =
...
+                      tv <- fresh
+                      let t = TArray tv
+                      attachTypeToPosition
+                        patLoc
+                        TypeMetadata
+                          { identExpr = patternToExpr $ bimap (const ()) (const ()) pattern,
+                            ty = (Set.empty, ImplType Map.empty $ t),
+                            docs = Nothing
+                          }
+                      return (t, [], Set.empty)
```
siddharth-krishna, line 1056: The empty array pattern `| [] ->` has type `array of 'a`

inferno-core/src/Inferno/Infer.hs
```diff
@@ -1031,17 +1034,50 @@ infer expr =
...
                       attachTypeToPosition patLoc meta
-                      return (t, [])
+                      return (t, [], Set.empty)
+                    PArray _ [] _ -> do
+                      tv <- fresh
+                      let t = TArray tv
+                      attachTypeToPosition
+                        patLoc
+                        TypeMetadata
+                          { identExpr = patternToExpr $ bimap (const ()) (const ()) pattern,
```
ngua, line 1052: One thing that might matter eventually, if `-XPatternSynonyms` is on (which it is in some `inferno` modules), this is going to break because `pattern` becomes a reserved keyword

inferno-core/src/Inferno/Infer.hs
```diff
@@ -988,13 +990,14 @@ infer expr =
...
+                `Set.union` (Set.unions patConstraints)
                 `Set.union` patTysEqConstraints
                 `Set.union` patTysMustEqCaseExprTy t_e
                 `Set.union` patExpTysEqConstraints (zip (map (\(_, ty, _) -> ty) res) (map (\(_, _p, _, e'') -> e'') patExprs))
                 `Set.union` (Set.unions cs_res)
             )
           where
-            mkPatConstraint :: Pat (Pinned VCObjectHash) SourcePos -> Infer (InfernoType, [(Ident, TypeMetadata TCScheme)])
+            mkPatConstraint :: Pat (Pinned VCObjectHash) SourcePos -> Infer (InfernoType, [(Ident, TypeMetadata TCScheme)], Set.Set Constraint)
             mkPatConstraint pattern =
```
ngua, line 1001: I guess this PR doesn't introduce the `pattern` thing, but I'd still recommend changing it, especially as some modules define pattern synonyms

inferno-core/src/Inferno/Infer/Exhaustiveness.hs
```diff
@@ -246,3 +249,20 @@ instance Enum EnumText where
...
+-- number of items in it.  For any pattern [<bla>], succ generates the pattern [<bla>, _].
+newtype EnumArrayPat = EnumArrayPat Int
+  deriving (Show)
+
+mkEnumArrayPat :: [(Pat a SourcePos, Maybe SourcePos)] -> EnumArrayPat
+mkEnumArrayPat = EnumArrayPat . length
+
+instance Pretty EnumArrayPat where
+  pretty (EnumArrayPat n) =
+    pretty $ PArray undefined (take n $ repeat $ (PVar (initialPos "") Nothing, Nothing)) undefined
```
ngua, line 264: Are the `undefined`s intentional? If so, maybe a comment would be helpful

inferno-core/src/Inferno/Module/Prelude/Defs.hs
```diff
@@ -355,6 +355,16 @@ minFun = bimap (min) (bimap (min) (min))
 maxFun :: Either3 Int64 Double EpochTime -> Either3 (Int64 -> Int64) (Double -> Double) (EpochTime -> EpochTime)
 maxFun = bimap (max) (bimap (max) (max))
 
+arrayIndexFun :: (MonadThrow m, Pretty c) => Value c m
+arrayIndexFun =
+  VFun $ \case
+    VArray a -> pure $ VFun $ \v -> do
+      i <- fromValue v
+      if 0 <= i && i < length a
```
ngua, line 363: Depending on the value of `i`, this could end up traversing the length of the list twice because `length` needs to traverse the whole list, and then `!!` also needs to traverse until at least `i`. I think it might be better to back the array type with a contiguous type, like a `Vector`, but that would be a much larger change of course, unrelated to this PR

Otherwise, you could use this [safe indexing function](https://hackage.haskell.org/package/extra-1.7.14/docs/src/Data.List.Extra.html#%21%3F) or write something equivalent to it to avoid traversing the list twice

inferno-core/src/Inferno/Infer/Exhaustiveness.hs
ngua: Also, `take n . repeat = replicate n`

albertov: I like `error "something searchable"` vs `undefined` for ease of debugging in case the impossible happens

siddharth-krishna:
Thanks @ngua -- I've done as you suggested, and also added a logging message as discussed with Scott so that we can monitor if people are doing a lot of slow random access, and if so we can move to using `Vector` in a future PR.

ngua (approved):

PR #57, opened 2023-07-22
Added note for no hasktorch support.

smurphy8:
Quick note about develop when hasktorch isn't supported.


smurphy8:
Hmm I spoke too soon, this doesn't give you cabal.

siddharth-krishna:
@ngua is there a way to get a shell without HaskTorch for inferno core work?

ngua:
It will require more hacks. I think that moving `inferno-ml` to its own repo might be better, although we'd have to keep it up to date with `inferno` of course

PR #58, opened 2023-07-24
Add `inferno-ml-remote`

ngua:
Adds a server that evaluates Inferno scripts with ML features (i.e. Hasktorch integration), with support for loading models and configuration for a model store and cache. At the moment, it only supports a single model store (so no support for user model stores)

This is used in [`inferno-ml-deploy`](https://github.com/plow-technologies/inferno-ml-deploy)

Note that this is largely a proof-of-concept at this point. An actual production system might be quite different


ngua:
The commit history is kind of messed up, I'm going to squash the changes onto a new branch and open a new PR

PR #59, opened 2023-07-24
Add `inferno-ml-remote`

ngua:
**Note**: Supersedes #58. Somehow the commit history got all messed up, so I just squashed the changes onto a new branch

Adds a server that evaluates Inferno scripts with ML features (i.e. Hasktorch integration), with support for loading models and configuration for a model store and cache. At the moment, it only supports a single model store (so no support for user model stores)

This is used in [`inferno-ml-deploy`](https://github.com/plow-technologies/inferno-ml-deploy)

Note that this is largely a proof-of-concept at this point. An actual production system might be quite different


inferno-ml-remote/app/Main.hs
```diff
@@ -0,0 +1,6 @@
+module Main (main) where
```
siddharth-krishna, line 1: Just a thought, would it make sense to put `inferno-ml-remote` in the same directory and cabal file as `inferno-ml` and just expose it as a separate exe?

inferno-ml-remote/test/Main.hs
```diff
@@ -0,0 +1,164 @@
...
+    astFromScript ::
+      FilePath -> IO (Either SomeInfernoError (Expr (Maybe VCObjectHash) SourcePos))
+    astFromScript = fmap (fmap snd . mkFinalAst) . readScript
+
+cacheModelsSpec :: Spec
+cacheModelsSpec =
+  Hspec.around withTempDir . Hspec.describe "cacheAndUseModel" $ do
+    Hspec.it "caches a model" $ \fp -> do
+      cacheAndUseModel "mnist.ts.pt" . mkCacheOption fp $ 10 * 1073741824
+      (`Hspec.shouldBe` ["mnist.ts.pt"]) =<< listDirectory fp
```
siddharth-krishna, line 129: In my `inferno-ml` PR, I was trying to avoid adding the ~4MB mnist model to the git repo. Perhaps I could modify the GitHub action to move this model file that it downloads to the right location for these tests?

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
```diff
@@ -0,0 +1,216 @@
...
+
+instance FromJSON Options where
+  parseJSON = withObject "Options" $ \o ->
+    Options <$> o .: "port" <*> o .:? "model-cache"
+
+newtype Script = Script Text
+  deriving stock (Show, Generic)
+  deriving newtype (Eq, FromJSON, ToJSON, IsString)
+
+newtype EvalResult = EvalResult Text
```
siddharth-krishna, line 112: Why is the eval result `Text`? Could we use `Value` and preserve the data types of the returned result?

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
```diff
@@ -0,0 +1,216 @@
...
+import Data.String (IsString)
+import Data.Text (Text)
+import qualified Data.Text as Text
+import Data.Word (Word64)
+import GHC.Generics (Generic)
+import qualified Options.Applicative as Options
+import Servant (Handler, JSON, Post, ReqBody, (:>))
+
+type InfernoMlRemoteAPI =
+  "inference" :> ReqBody '[JSON] Script :> Post '[JSON] EvalResult
```
siddharth-krishna, line 46: The input type could/should be `Expr` instead of `Text`, as I'm guessing scripts will be saved in inferno-vc, which stores scripts as the `Expr` ASTs instead of as text. That way you can also avoid parsing the scripts for every request.

siddharth-krishna (commented):
Thanks Rory, this is awesome! I had a few questions/comments.

inferno-ml-remote/test/Main.hs
ngua: I understand the desire behind that, but that means that this whole test won't be buildable with Nix (unless I reorganized it) and I'm also using it in some `nixosTest`s that require no network access (it needs to be pure)

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
ngua: There's no `ToJSON` instance. If you think we can write an instance that makes sense, I'd also prefer that

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
ngua: A similar thing applies to the comment above. There's no `FromJSON` instance, but if we can write a lawful, sensible instance then I'd also prefer that

inferno-ml-remote/test/Main.hs
siddharth-krishna: I see. Is it okay to add hasktorch as a dependency? Then I can provide a Haskell function that creates a random ML model of whatever size you require and save it as a `.ts.pt` file. Would that work?

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
siddharth-krishna: I think we do have one, don't we? `inferno-vc` stores `Expr`s as JSON. Does this not work?
https://github.com/plow-technologies/inferno/blob/2ca4041cb4c9dfe611b051edde59430ea6ad1942/inferno-types/src/Inferno/Types/Syntax.hs#L719

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
siddharth-krishna: I think we should be able to write a sensible `ToJSON`. Perhaps even the default/deriving one might be okay?

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
ngua: Oh I must have missed that. We'd have to change how several things currently work. I'd prefer to merge this first and then figure that out

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
ngua: I don't think we can derive an instance because one of the constructors holds a function (`VFun`)

inferno-ml-remote/test/Main.hs
siddharth-krishna: @ngua maybe you missed this comment?

inferno-ml-remote/src/Inferno/ML/Remote/Types.hs
siddharth-krishna: Oh yes, good catch. But that's for intermediate interpreter values only. The final result of a script will never be a function, and we check this when we type check scripts:
https://github.com/plow-technologies/inferno/blob/a143f90433bf12b3115edf4e2f7ab698c87187a2/inferno-lsp/src/Inferno/LSP/ParseInfer.hs#L538-L542

So perhaps we can serialize `VFun` to some suitable error value /  raise an exception?

PR #60, opened 2023-07-26
Upgrade to GHC 9.2.5

### LINKED ISSUE #25: Upgrade to GHC 9.2.5
#20 bumped the haskell.nix revision. It seems that 9.2.5 will be the new supported 9.2.x compiler version. I've noticed at least one cache miss with 9.2.4: `devShells.<SYSTEM>.default` now requires rebuilding GHC itself if `shell.tools.haskell-language-server` is configured. I've confirmed that bumping the version to 9.2.5 avoids this (although there are other unrelated build issues with HLS plugins). I disabled HLS in #24, but we will probably see more cache misses in the future if we don't upgrade from 9.2.4.

ngua:
Closes #25 

I'd like to do this for a few reason:
- `haskell-language-server` integration is currently broken, this fixes that
- we're planning GHC upgrades elsewhere, it's a good idea to do this incrementally, especially considering our recent Hasktorch integration
- using `cabal repl` with GHC 9.2.4 (which is my primary development mode) has an irritating bug where auto-completing any module name leads to a compiler panic

Note: I haven't changed anything else regarding e.g. the `index-state`, so we're still using the same dependencies. The only other change is that now some of the flake outputs are suffixed with `-ghc925` instead of `-ghc924`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,7 @@ jobs:
           extra_nix_config: |
             substituters = https://cache.nixos.org https://cache.iog.io
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
+            narinfo-cache-negative-ttl = 60
       - uses: cachix/cachix-action@v12
         with:
           name: inferno
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
           extra_nix_config: |
             substituters = https://cache.nixos.org https://cache.iog.io
             trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ=
-      - uses: cachix/cachix-action@v8
+      - uses: cachix/cachix-action@v12
         with:
           name: inferno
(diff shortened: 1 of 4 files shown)


siddharth-krishna (commented):
I'm happy to upgrade! My only concern is that for the interim period when this repo uses 9.2.5 and `all` uses 9.2.4 could some build bugs slip through the CI?

Perhaps there's some way to switch to 9.2.5 for all practical purposes but keep a 9.2.4 build test in CI until we also upgrade `all`?

ngua:
@siddharth-krishna I don't think there should be any significant bugs related to changing the minor version. I'd say the more important thing is keeping the versions of dependencies the same, which this does. But if you're very concerned about that, I could try adding another set of things for 9.2.4. My only concern is that evaluating flake outputs is going to get slower because that requires re-evaluating the whole haskell.nix project (at least some of it gets evaluated every time you build something, regardless of which compiler you're building for)

siddharth-krishna:
Okay, if build bugs aren't a real concern between minor versions I'm happy to upgrade and live on the edge in order to have shorter build and nix times. :)

ngua:
Hmmm. This is really bizarre. CI is building GHC from source. It works fine locally

ngua:
Now I'm _really_ confused. I thought maybe the haskell.nix cache no longer cached the version of GHC we need. I had tried upgrading a while ago and haven't done a GC since, so I thought maybe that explained why I'm not running into this

I pushed the paths required to build `packages..default` to our cache directly, and yet it's _still_ building GHC

ngua:
I think I know the issue, doing some tests now

ngua:
@siddharth-krishna If you have the time at some point, could you try running `nix develop` on this branch please and let me know if you end up building GHC 9.2.5 from source at any point?

siddharth-krishna:
`nix build` was very quick, but looks like `nix develop` is still building GHC:
image

ngua:
@siddharth-krishna I ran the CI job again and it's working fine now. I'm going to merge this, if we have issues later we can just revert it

PR #66, opened 2023-08-04
Fix pattern match eval of enums

### LINKED ISSUE #64: Boolean pattern match evaluates incorrectly
```ocaml
match 1 < 2 with {
  | #false -> 1.1
  | #true -> 0.0
}
```
Evaluates to `1.1` but it should be `0.0`.

siddharth-krishna:
Fixes #64


goodlyrottenapple (approved):
thanks!!

PR #67, opened 2023-08-04
Fix `torch` versions, document, use on CI

### LINKED ISSUE #65: Torch version mismatch
As far as I can tell, Nix is not grabbing the version of pytorch specified here: https://github.com/plow-technologies/inferno/blob/8d1a5eee65fab73afc28c185199b44ece9b97b30/.github/workflows/build.yml#L50-L51.

ngua:
Closes #65. Previously `devShells..pytorch` was using an incorrect version (one major version too high). I added something to the docs mentioning `devShells..pytorch` and also used it to run the MNIST example on CI. This way we will always be using the same version everywhere. I temporarily modified the workflow file in order to ignore the cache and ran the CI job manually and it worked (otherwise that part wouldn't run because it's been cached)

### DIFF
--- a/.github/torch.sh
+++ b/.github/torch.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+
+mkdir -p "$(dirname "${MNIST_FNAME}")"
+cd "$(dirname "${MNIST_FNAME}")" || exit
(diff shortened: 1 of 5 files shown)


mirko-plowtech (approved):
LGTM!

PR #69, opened 2023-08-04
Generate `cabal.project.local` in the `shellHook`

ngua:
This removes the awful hack I originally implemented to get the conditional import in the `cabal.project` working with haskell.nix. The GHC 9 config is still read in conditionally in the haskell.nix setup, but I removed it from the main `cabal.project` (so it doesn't need to be removed when building with Nix as before)

For building outside of Nix, the `shellHook` in `devShells..default` now generates a `cabal.project.local` with the `import` stanza (but it won't overwrite an existing file). This is nice because it can imported using an absolute path this, which wasn't possible before, so Cabal and HLS will actually work now if invoked from a different directory than the project root


mirko-plowtech (approved):
LGTM!

PR #77, opened 2023-09-01
[inferno-vc] Simplified InfernoVCOperations typeclass usage

albertov:
Also removed `getAllHeads` and added `deleteAutosavedVCObjectsOlderThan` instead to allow for more efficient implementations.

The changes to the class that make the usage simpler are:

 - Removed the `De/Serializable` associated constraints which were applied on the `group` and `author` type variables of `VCMeta` and added a `Group` and `Author` associated types instead. See the changes in `runServer` and `runServerConfig` call sites.
  This change is not only cosmetic, it also gives knowledge of `author` and `group` to all type-class method which can use it to deserialise stuff without resorting to AmbiguousTypes or Proxies (which moves the burden of keeping track of the type witnesses to the call-sites)

- Made the constraint `AsType err VCStoreError` that `runServer/Config` requires explicit on the type-class. This is so implementors don't assume they can ignore it until they're faced with it when attempting to make the call-sites compile.

Removing `getAllHeads` is motivated by the fact that the only caller was `Inferno.VersionControl.deleteStaleAutosavedVCObjects`. This function used an implementation that can be made much more efficient if the backend knows that what is actually meant is "give me all the heads _so I can purge some auto-saved objects_" so it has been moved to the typeclass in its place

### DIFF
--- a/inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
+++ b/inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
@@ -153,11 +153,11 @@ instance
               let new_head_fp = storePath </> "heads" </> show obj_h
               liftIO $ renameFile head_fp new_head_fp
               -- we append the previous head hash to the file (this serves as lookup for all the predecessors)
-              appendBS new_head_fp $ BL.fromStrict $ vcObjectHashToByteString pred_hash <> "\n"
+              appendBS new_head_fp $ vcObjectHashToByteString pred_hash <> "\n"
               -- now we need to change all the predecessor mappings in '<storePath>/to_head' to point to the new HEAD
               -- we also include the new head pointing to itself
               preds <- readVCObjectHashTxt new_head_fp
-              let obj_h_bs = BL.fromStrict $ vcObjectHashToByteString obj_h
+              let obj_h_bs = vcObjectHashToByteString obj_h
               forM_ (obj_h : preds) (\pred_h -> writeBS (storePath </> "to_head" </> show pred_h) obj_h_bs)
 
               pure obj_h
@@ -169,7 +169,7 @@ instance
           appendBS new_head_fp mempty
 
           -- we again make sure to add a self reference link to the '<storePath>/to_head' map
-          let obj_h_bs = BL.fromStrict $ vcObjectHashToByteString obj_h
+          let obj_h_bs = vcObjectHashToByteString obj_h
           writeBS (storePath </> "to_head" </> show obj_h) obj_h_bs
           pure obj_h
 
@@ -178,7 +178,7 @@ instance
       writeBS (storePath </> "deps" </> show obj_h) mempty
       forM_ deps $ \dep_h -> do
         -- first we append the direct dependency hash 'dep_h'
-        appendBS (storePath </> "deps" </> show obj_h) $ BL.fromStrict $ vcObjectHashToByteString dep_h <> "\n"
+        appendBS (storePath </> "deps" </> show obj_h) $ vcObjectHashToByteString dep_h <> "\n"
         -- then we append the transitive dependencies of the given object, pointed to by the hash 'dep_h'
         appendBS (storePath </> "deps" </> show obj_h) =<< getDepsFromStore (storePath </> "deps") dep_h
 
@@ -431,21 +431,21 @@ checkPathExists fp =
     False -> throwError $ CouldNotFindPath fp
     True -> pure ()
 
-getDepsFromStore :: (VCStoreLogM env m, VCStoreErrM err m) => FilePath -> VCObjectHash -> m BL.ByteString
+getDepsFromStore :: (VCStoreLogM env m, VCStoreErrM err m) => FilePath -> VCObjectHash -> m B.ByteString
 getDepsFromStore path h = do
   let fp = path </> show h
   checkPathExists fp
-  liftIO $ BL.readFile $ path </> show h
+  liftIO $ B.readFile $ path </> show h
 
-appendBS :: (VCStoreLogM env m) => FilePath -> BL.ByteString -> m ()
+appendBS :: (VCStoreLogM env m) => FilePath -> B.ByteString -> m ()
 appendBS fp bs = do
   trace $ WriteTxt fp
-  liftIO $ BL.appendFile fp bs
+  liftIO $ B.appendFile fp bs
 
-writeBS :: (VCStoreLogM env m) => FilePath -> BL.ByteString -> m ()
+writeBS :: (VCStoreLogM env m) => FilePath -> B.ByteString -> m ()
 writeBS fp bs = do
   trace $ WriteTxt fp
-  liftIO $ BL.writeFile fp bs
+  liftIO $ B.writeFile fp bs
 
 writeHashedJSON :: (VCStoreLogM env m, VCHashUpdate obj, ToJSON obj) => FilePath -> obj -> m VCObjectHash
 writeHashedJSON path o = do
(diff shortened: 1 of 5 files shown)


inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
```diff
@@ -177,7 +191,7 @@ instance (MonadIO m, MonadMask m, AsType VCStoreError err) => InfernoVCOperation
     lock <- asks getTyped
     withWrite lock $ do
       -- check if object meta exists with hash meta_hash, and get meta
-      (VCMeta {name = obj_name} :: VCMeta Value Value VCObject) <- fetchVCObject obj_hash
+      VCMeta {name = obj_name} <- fetchVCObject obj_hash
```
albertov, line 194: This change is made possible because `deleteAutosavedVCObject` now knows what `g` and `a` are so we don't have to "upcast"

inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
```diff
@@ -204,7 +218,7 @@ instance (MonadIO m, MonadMask m, AsType VCStoreError err) => InfernoVCOperation
       createDirectoryIfMissing True $ storePath </> "removed" </> "deps"
 
     withWrite lock $ do
-      (metas :: [VCMeta Value Value VCObjectHash]) <- fetchVCObjectHistory obj_hash
+      metas <- fetchVCObjectHistory obj_hash
```
albertov, line 221: ditto This change is made possible because deleteAutosavedVCObject now knows what g and a are

inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
```diff
@@ -316,20 +330,35 @@ instance (MonadIO m, MonadMask m, AsType VCStoreError err) => InfernoVCOperation
...
-    pure $
-      foldr
-        ( \hd xs ->
-            case (either (const Nothing) Just $ Base64.decode $ Char8.pack hd) >>= digestFromByteString of
-              Just hsh -> (VCObjectHash hsh) : xs
-              Nothing -> xs
-        )
-        []
-        (map takeFileName headsRaw)
+  deleteAutosavedVCObjectsOlderThan t = do
```
albertov, line 333: sqlite backend can now do this with an efficient `DELETE FROM autosaved_objects WHERE timestamp < ?`

inferno-vc/src/Inferno/VersionControl/Server.hs
```diff
@@ -125,58 +120,45 @@ vcServer toHandler =
...
+    VCHashUpdate (Ops.Author m),
+    VCHashUpdate (Ops.Group m),
+    FromJSON config,
+    FromJSON (Ops.Author m),
+    FromJSON (Ops.Group m),
+    ToJSON (Ops.Author m),
+    ToJSON (Ops.Group m),
+    Ops.InfernoVCOperations VCServerError m
   ) =>
-  proxy a ->
```
albertov, line 173: bye bye proxies

inferno-vc/src/Inferno/VersionControl/Server.hs
```diff
@@ -185,8 +167,8 @@ runServerConfig _ _ withEnv runOp serverConfig = do
       serverTracer = contramap vcServerTraceToText tracer
   withEnv serverConfig tracer $ \env -> do
     let cleanup = do
-          now <- liftIO $ CTime . round . toRational <$> getPOSIXTime
-          runExceptT (runOp (deleteStaleAutosavedVCObjects @a @g now) env) >>= \case
+          now <- liftIO getPOSIXTime
+          runExceptT (runOp (Ops.deleteAutosavedVCObjectsOlderThan now) env) >>= \case
```
albertov, line 171: bye type witness

inferno-vc/test/Spec.hs
```diff
@@ -22,10 +21,8 @@ main =
     _ <-
       forkIO $
         runServerConfig
-          (Proxy :: Proxy Int)
```
albertov, line 25: bye proxies

inferno-vc/test/Spec.hs
```diff
@@ -22,10 +21,8 @@ main =
...
-          (Proxy :: Proxy Int)
           FSOps.withEnv
-          FSOps.runInfernoVCFilesystemM
+          (FSOps.runInfernoVCFilesystemM @Int @Int)
```
albertov, line 25: hello TypeApplications  (alternatively a scoped type signature with ScopedTypeVariables)

inferno-vc/src/Inferno/VersionControl/Server.hs
ngua: "Bye bye proxies" sounds like a Haskell nursery rhyme

inferno-vc/src/Inferno/VersionControl/Server.hs
```diff
@@ -125,58 +120,45 @@ vcServer toHandler =
...
-    Ops.Deserializable m g,
-    Ops.Serializable m a,
-    Ops.Serializable m g,
-    Ops.InfernoVCOperations VCServerError m,
-    Ord g
+    FromJSON (Ops.Author m),
+    FromJSON (Ops.Group m),
+    ToJSON (Ops.Author m),
+    ToJSON (Ops.Group m),
+    Ops.InfernoVCOperations VCServerError m
```
albertov, line 133: we're stuck with this constraint and VCServerError is not extensible by implementors so made this constraint explicit on the typeclass context

inferno-vc/src/Inferno/VersionControl/Server.hs
```diff
@@ -125,58 +120,45 @@ vcServer toHandler =
...
-runServer proxyA proxyG withEnv runOp = do
+runServer withEnv runOp = do
   readServerConfig "config.yml" >>= \case
     Left err -> putStrLn err
-    Right serverConfig -> runServerConfig proxyA proxyG withEnv runOp serverConfig
+    Right serverConfig -> runServerConfig withEnv runOp serverConfig
 
 runServerConfig ::
-  forall proxy m a g env config.
+  forall m env config.
```
albertov, line 144: `a` and `g` can now be inferred from `m` thanks to the `InfernoVCOperations` constraint so we don't need no proxies

inferno-vc/src/Inferno/VersionControl/Server.hs
siddharth-krishna: Shouldn’t it be `now - 60*60`?

siddharth-krishna (commented):
Very nice! Potential bug: did we lose the constraint of deleting auto saves from before 1 hour ago?

inferno-vc/src/Inferno/VersionControl/Server.hs
albertov: Right. Good catch!

siddharth-krishna (approved):

inferno-vc/src/Inferno/VersionControl/Operations/Filesystem.hs
```diff
@@ -204,8 +218,8 @@ instance (MonadIO m, MonadMask m, AsType VCStoreError err) => InfernoVCOperation
       createDirectoryIfMissing True $ storePath </> "removed" </> "deps"
 
     withWrite lock $ do
-      (metas :: [VCMeta Value Value VCObjectHash]) <- fetchVCObjectHistory obj_hash
-      forM_ metas $ \VCMeta {obj = hash} -> do
+      metas <- fetchVCObjectHistory obj_hash
+      forM_ (takeUpUntilClone metas) $ \VCMeta {obj = hash} -> do
```
albertov, line 222: This fixes original object being deleted when clone chain was

Issue #78, opened 2023-09-01
Figure out caching problem


ngua:
Ever since the GHC 9.2.5 upgrade there have been very frequent cache misses. I don't think this is an issue with Cachix; I've checked open issues for Cachix (and its GH action) and can't find anything related. Cachix is very widely used so if there were an issue with this I imagine it would have been reported by now

We might be getting IOG cache misses. I'm going to try to bump the haskell.nix input and make some other changes to see if I can get it cached reliably

ngua:
I fixed this a while ago

PR #79, opened 2023-09-06
Add magic cache

ngua:
I was working on an ad-hoc solution for #78 and then I found that there's an existing GH action that does basically the same thing. I've triggered several (expected) rebuilds on a different branch with the same actions config and haven't experienced any rebuilds of dependencies. It appears to work fine with Cachix as well

One advantage to also using this is that if an external contributor ever wanted to open a PR, the Cachix cache wouldn't work because it always requires a secret, which are unavailable for PRs from forks

Note that the `magic-cache-action` uses the built-in GH cache as a substituter (rather than mounting a cache as the Nix store), so dependencies will still be downloaded in any event

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
         with:
           name: inferno
           authToken: "${{ secrets.CACHIX_TOKEN }}"
+      - uses: DeterminateSystems/magic-nix-cache-action@main
       - run: |
           nix build -L .#
 


siddharth-krishna (approved):
Nice find! Hopefully, this is the end of our cache troubles :)

ngua:
@siddharth-krishna I wasn't able to reproduce any dependency rebuilds on another branch with the same setup (just normal rebuilds of `inferno` things). If this doesn't work for some reason, I will make a similar cache setup but mount it as the Nix store instead of using this action

siddharth-krishna:
Sounds great, thanks!

PR #86, opened 2023-10-03
Run Interpreter in custom monad `m`

siddharth-krishna:
This PR generalizes the `Interpreter` API to run in a user supplied monad `m` instead of `IO`. This can be used to run the interpreter in some context, see the `inferno-core/test/Eval/Spec.hs` test for an example.


albertov (approved):
Looks good 💪

PR #89, opened 2023-10-11
Add type annotations to let expressions

siddharth-krishna:
This PR allows adding type annotations to let declarations:
```ocaml
// Type error:
let t : int -> int = truncate in t 43.3

// OK. y should have type int
let x : int = 3 in let y = truncate x in y
```


inferno-types/src/Inferno/Types/Syntax.hs
```diff
@@ -89,6 +89,30 @@ module Inferno.Types.Syntax
...
+    typeInt,
+    typeBool,
+    typeDouble,
+    typeWord16,
+    typeWord32,
+    typeWord64,
+    typeText,
+    typeResolution,
+    typeTimeDiff,
+    typeTime,
```
siddharth-krishna, line 115: All these type declarations had to move into this file from `Type.hs` because `Expr` now depends on `TCScheme` (because the new `LetAnnot` constructor holds a type annotation). Is that okay?

inferno-types/src/Inferno/Types/Type.hs
```diff
@@ -35,206 +35,33 @@ module Inferno.Types.Type
...
-
-data TypeMetadata ty = TypeMetadata
-  { identExpr :: Expr () (),
-    docs :: Maybe Text,
-    ty :: ty
-  }
-  deriving (Eq, Show, Generic, Data, ToJSON, FromJSON)
+import Inferno.Types.Syntax
+  ( BaseType (..),
+    ImplType (..),
```
siddharth-krishna, line 40: I've left this module as is, in case there was some reason for separating these type definitions from the others in `Syntax.hs`. If there's no reason to keep this separate file, I can remove it.

goodlyrottenapple: Syntax.hs is already pretty huge so might be good to keep some bits of it, like the types, separate

inferno-core/src/Inferno/Eval.hs
```diff
@@ -169,6 +169,11 @@ eval env@(localEnv, pinnedEnv) expr = case expr of
         argv <- eval env arg
         f argv
       _ -> throwM $ RuntimeError "failed to match with a function"
+  -- TODO create pattern for LetAnnot, or reuse?
```
goodlyrottenapple, line 172: yeah `LetAnnot_` might be nice

inferno-core/src/Inferno/Infer.hs
```diff
@@ -784,23 +784,57 @@ infer expr =
...
+          (e2', ImplType i2 t2, c2) <- inEnv newEnv $ infer e2
+          let (isMerged, ics) = mergeImplicitMaps (blockPosition expr) [i1, i2, iT]
+          return
+            ( LetAnnot p1 loc x pT t p2 e1' p3 e2',
+              ImplType isMerged t2,
+              Set.fromList ics
+                `Set.union` c1
+                `Set.union` c2
+                -- Type of e1 == type annotation
+                `Set.union` Set.fromList [tyConstr t1 tT [UnificationFail tyCls t1 tT $ blockPosition e1]]
```
goodlyrottenapple, line 817: might be worth adding a new error message `UnificationFailAnnot` which gives a more detailed message, i.e. something along the lines of "you said x is of type t but I inferred the RHS looks to be of type y"

inferno-core/src/Inferno/Eval.hs
siddharth-krishna: I was wondering if there's a way to reuse the `Let_` pattern for both cases since the eval code is the same for either case as it ignores annotations?

inferno-types/src/Inferno/Types/Type.hs
siddharth-krishna: Thanks, but since `Expr` now depends on `TCScheme`, there's going to be cyclic dependencies.. unless I try to reverse the dependency and have `Syntax.hs` import `Type.hs`? Some things like `TypeMetadata` might have to stay in `Syntax.hs` since it depends on `Expr`.

inferno-core/src/Inferno/Infer.hs
```diff
@@ -784,23 +784,57 @@ infer expr =
...
+                -- Type class constraints from type annotation TODO filter out reps?
+                `Set.union` Set.map (Right . (exprLoc,)) tcs
```
siddharth-krishna, line 819: @goodlyrottenapple I saw that the code for `Var` was [filtering out](https://github.com/plow-technologies/inferno/blob/af58127ad12758185d78c313ff911d873a9e90bb/inferno-core/src/Inferno/Infer.hs#L562) the `rep` typeclass constraints, why does it do that and would we need to do something similar here?

inferno-types/src/Inferno/Types/Type.hs
goodlyrottenapple: ah good point... dunno may be worth making submodules for Expr.<...> for the auxiliary stuff like TypeMetadata...  I'll leave it up to you

inferno-core/src/Inferno/Infer.hs
goodlyrottenapple: The rep typeclass doesn't really constrain the type in any way, because we have an an instance for every type, so it was easier to just remove it than pass it to the constraint solver and deal with it there. As far as i remember, the inference modifies a type-checked term and inserts `TyRep` wherever we need the runtime information, so it should be safe to remove the ty-rep typeclass information at call sites of already type-checked terms?

As to the call site here, you shouldn't allow the user to write a tyrep typeclass in the first place, so probably no need to try to filter it out?

inferno-core/src/Inferno/Eval.hs
goodlyrottenapple: Don't think that's possible...

inferno-types/src/Inferno/Types/Type.hs
siddharth-krishna: Okay thanks, I'll leave the restructuring for a future PR then. If no objections, I'll merge this PR in now?

PR #94, opened 2023-11-07
Chain prelude evaluation so that primitives can depend on each other

siddharth-krishna:
This PR allows evaluating a second prelude (like the one in `inferno-ml`) in an environment that already has the first prelude (the core/builtin one), so that the second prelude can use ops and primitives defined in the first one.

It creates a new `Prelude` type (that subsumes the type alias `ModuleMap`) and refactors existing code to use it.


inferno-core/inferno-core.cabal
```diff
@@ -38,7 +38,6 @@ library
     , Inferno.Parse.Commented
     , Inferno.Parse.Error
     , Inferno.Instances.Arbitrary
-    , Inferno.Utils.QQ.Script
```
siddharth-krishna, line 41: Removed this module as we don't use it anywhere. An appendix from the old frontend perhaps?

inferno-core/src/Inferno/Module.hs
```diff
@@ -1,15 +1,24 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Inferno.Module
   ( Module (..),
     PinnedModule,
+    Prelude (..),
```
siddharth-krishna, line 9: This file is now basically functions on the `Prelude` type. Should we rename it to `Inferno.Prelude`? But then what do we call `Inferno.Module.Prelude`, which has the actual builtin prelude definition? (I don't think it can be the same file because TH needs the QuasiQuoter defined in a separate module)

inferno-core/test/Eval/Spec.hs
```diff
@@ -426,25 +426,27 @@ cachedGet =
...
 
 evalInMonadTest :: Spec
 evalInMonadTest = do
   let testEnv = TestEnv {cache = 4}
 
-  let modules =
-        Map.unionWith
-          (error "Duplicate module name in builtinModules")
-          (Prelude.builtinModules @(ReaderT TestEnv IO) @TestCustomValue)
-          evalInMonadPrelude
```
siddharth-krishna, line 445: The nice thing about chaining prelude evaluation is that we don't have to combine the preludes in a hacky way like this anymore

Daniel-Diaz:
Looks like a nice improvement!

Daniel-Diaz (approved):

PR #97, opened 2024-01-09
[inferno-vc] Add tests for fetchFunctionsForGroups

siddharth-krishna:
This PR adds some tests that would have caught the recent bugs we had with the Cardith backend. (I tested manually that these tests fail before the bugfix commit.)


albertov (approved):
Lgtm!

PR #102, opened 2024-02-26
Add `inferno-ml-server-types` package

ngua:
Adds the `inferno-ml-server-types` package containing the API for a server that can evaluate Inferno scripts using ML primitives. There is no actual implementation of `inferno-ml-server` here

This and the whole Inferno ML infrastructure is still a WIP. However, a WIP implementation will need to be merged here in order to merge some WIPs in other repos. The following at least are not yet complete:
- making a polymorphic prelude for `inferno-ml` that doesn't depend on `hasktorch` (helpful for typechecking in cases where we can't use `hasktorch`)
- golden and other tests for the new types; however, since many of the types are polymorphic on something, it probably makes more sense to just add them in one of the places where `inferno-ml-server` or the bridge server are implemented

I've also made a few changes to `inferno-ml`, which are listed in the changelog


inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -0,0 +1,762 @@
...
+  "status" :> Get '[JSON] (Maybe ())
+    -- Evaluate an inference script
+    :<|> "inference"
+      :> Capture "id" (Id (InferenceParam uid gid p s))
+      :> QueryParam "res" Int64
+      :> Post '[JSON] ()
+    :<|> "inference" :> "cancel" :> Put '[JSON] ()
+    -- Register the bridge. This is an `inferno-ml-server` endpoint, not a
+    -- bridge endpoint
+    :<|> "bridge" :> ReqBody '[JSON] BridgeInfo :> Post '[JSON] ()
```
ngua, line 111: We might want to just embed the bridge info directly into the servers (if we use a stable domain name, for example). But doing it this way allows for the bridge info to be updated dynamically (i.e. while the `inferno-ml-server` is running)

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -0,0 +1,762 @@
...
+    where
+      fromChars :: (Char, Char) -> Maybe ISO63912
+      fromChars c = ISO63912 c <$> Map.lookup c languagesByCode
+
+instance ToJSON ISO63912 where
+  toJSON (ISO63912 (f, s) _) = String . flip Text.snoc s $ Text.singleton f
+
+{- ORMOLU_DISABLE -}
+languagesByCode :: Map (Char, Char) Text
+languagesByCode =
```
ngua, line 697: I don't know if we actually need these, but it might be nice for an eventual frontend to only display valid ISO 6391-2 codes. (There's a package for this on hackage, but it's ancient and uses a type with 100+ constructors to represent the code)

inferno-ml/src/Inferno/ML/Types/Value.hs
```diff
@@ -9,30 +9,33 @@ import Language.Haskell.TH.Quote (QuasiQuoter (..))
 import Prettyprinter (Pretty (pretty), align)
 import qualified Torch as T
 
-data MlValue
+data MlValue x
   = VTensor T.Tensor
   | VModel T.ScriptModule
+  | VExtended x
```
ngua, line 15: This is pretty awful and hacky, but I can't think of a better way to allow additional types (which is necessary in my WIP `inferno-ml-server` implementation). The problem is that `VCustom` is set to `MlValue`, but we need the custom type to be able to represent additional things (not just `MlValue`s). I guess it would be possible to create another type that can hold an `MlValue` or another type, but then it would be necessary to map over all of the `ML` prelude functions. Or perhaps it might be possible once I add the more polymorphic `ML` prelude

If anyone has a better idea, please let me know!

inferno-ml-server-types/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision History for inferno-ml-server-types
+*Note*: we use https://pvp.haskell.org/ (MAJOR.MAJOR.MINOR.PATCH)
+
+## 0.0.1
+* Initial pre-release
```
ngua, line 5: I would prefer to include a textual tag indicating a pre-release, but Cabal doesn't like that

inferno-ml/src/Inferno/ML/Types/Value.hs
siddharth-krishna: Yeah I knew this would be a problem when I introduced `VCustom`... Is there a way to use a typeclass to make this nicer? E.g.
```haskell
zerosFun :: (MonadThrow m, CustomValue c, HasTensorValue c) => Value c m
```

siddharth-krishna (commented):
Partial review: reviewed changes to inferno-ml, now looking through inferno-ml-server-types

inferno-ml/src/Inferno/ML/Types/Value.hs
ngua: Yes, that could work. Maybe something like

```haskell
class HasMlTypes c where
  type Tensor c
  type Model c
  ... 
```

would work for both of the required types

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -0,0 +1,762 @@
...
+-- script type
+data InferenceParam uid gid p s = InferenceParam
+  { id :: Maybe (Id (InferenceParam uid gid p s)),
+    -- | The script of the parameter
+    --
+    -- For new parameters, this will be textual or some other identifier
+    -- (e.g. a UUID for use with @inferno-lsp@)
+    --
+    -- For existing inference params, this is the foreign key for the specific
+    -- script in the 'InferenceScript' table
```
siddharth-krishna, line 499: Curious about this -- it looks like it's for backwards compatibility but we haven't yet launched inferno-ml -- is this something we intend to remove before the public launch?

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -0,0 +1,762 @@
...
+-- by @inferno-ml-server@ or the bridge server
+--
+-- Note that this is significantly more restrictive than Inferno's @Value@ type,
+-- which cannot have sensible @ToJSON@ and @FromJSON@ instances
+data IValue
+  = IText Text
+  | IDouble Double
+  | ITuple (IValue, IValue)
+  | ITime EpochTime
+  | IEmpty
```
siddharth-krishna, line 607: We might also want booleans. Also, perhaps we should sync this with the inferno-tachdb value conversions in plow-inferno, to make sure both support the same set of value types

inferno-ml-server-types/src/Inferno/ML/Server/Client.hs
```diff
@@ -0,0 +1,43 @@
...
+where
+
+import Data.Int (Int64)
+import Data.Proxy (Proxy (Proxy))
+import Inferno.ML.Server.Types
+import Servant ((:<|>) ((:<|>)))
+import Servant.Client.Streaming (ClientM, client)
+
+-- | Get the status of the server. @Nothing@ indicates that an inference job
+-- is being evaluated. @Just ()@ means the server is idle
```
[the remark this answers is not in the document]
ngua: Yes, I agree it's counterintuitive. It's just from my laziness dealing with the `MVar` in the actual implementation (if the `MVar` is empty or not). I could swap it

inferno-ml-server-types/src/Inferno/ML/Server/Client.hs
```diff
@@ -0,0 +1,43 @@
...
+statusC :: ClientM (Maybe ())
+
+-- | Run an inference parameter
+inferenceC :: Id (InferenceParam uid gid p s) -> Maybe Int64 -> ClientM ()
```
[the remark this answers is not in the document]
ngua: Yes, I should document that

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
ngua: Yes, I agree. We should probably add a type representing a limited subset of Inferno values to one of the OSS packages and use that in both places

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
ngua: Actually, it's so that we can support posting all of the information to create an inference param without needing to have a `script` field be a `VCObjectHash`. So for example, it could be the UUID of a script parsed by `inferno-lsp` 

By "existing" param, I mean when we save a param to the DB, this field will be a `VCObjectHash`, not that there are params saved somewhere already that we need to maintain backwards compatibility with. I will rephrase it to make that clearer

ngua:
Thanks @siddharth-krishna, I will address some of the comments/ideas as I keep working on this

PR #103, opened 2024-02-28
Add records to the Inferno language

siddharth-krishna:
This PR brings records to the Inferno language.

### Syntax:
```ocaml
let r1 = {ht = 1.51; wt = 62.4} in
let bmi = fun r -> r.wt / (r.ht * r.ht) in
bmi r1
```

### Types:
A general record type looks like `{f1: t1; ...; fN: tN; 𝜌}` where 𝜌 is a type variable called a *row variable* representing either:
- `⏊`: a special symbol denoting no further fields in the record, or
- one or more fields with types `f1': t1'; ...; fN': tN'`

The example above has the following types:
```ocaml
r1 : {ht: double; wt: double}
bmi : forall 'a. {ht: double; wt: double; 'a} -> double
```
Here, the concrete record `r1` has type `{ht: double; wt: double}` (which is syntactic sugar for `{ht: double; wt: double; ⏊}`, the record having exactly two fields), and the type of the function `bmi` uses a row variable `'a` denotes the fact that it accepts as argument any record that has at least the fields `ht` and `wt`.

### Differences to OCaml's Immediate Objects

The description above is very similar to OCaml's ad-hoc objects, except that we retain the row variables in the types and show it to the user. The example above in OCaml would have the types:
```ocaml
r1 : {ht: double; wt: double}
bmi : {ht: double; wt: double; ...} -> double
```

I decided to keep the row variables explicit for now, because it makes the quantification on the "rest of the record fields" more explicit when we manually annotate functions. We can always change the pretty-printer and parser later to be like OCaml and hide row variables using the `...` syntax, if we prefer.


inferno-core/src/Inferno/Infer.hs
```diff
@@ -379,7 +388,7 @@ inferTypeReps allTypeClasses (ForallTC tvs tyCls (ImplType _impl ty)) inputTys o
   let cs =
         [Right (dummyPos, c) | c@(TypeClass nm _) <- Set.toList tyCls, nm /= "rep"]
           ++ mkConstraints ty inputTys
-   in case runSolve allTypeClasses cs of
+   in case runSolve 0 allTypeClasses cs of -- TODO what if 't0 is in the type passed as argument to this function?
```
siddharth-krishna, line 391: @goodlyrottenapple would I need to find the largest type var in the constraints `cs` and set the initial `varCount` to that plus one?

inferno-core/src/Inferno/Infer.hs
```diff
@@ -1206,20 +1257,128 @@ compose :: Subst -> Subst -> Subst
...
 -- | Run the constraint solver
-runSolve :: Set.Set TypeClass -> [Constraint] -> Either [TypeError SourcePos] Subst
-runSolve allClasses cs = runIdentity $ runExceptT $ flip runReaderT allClasses $ solver st
+runSolve :: Int -> Set.Set TypeClass -> [Constraint] -> Either [TypeError SourcePos] Subst
+runSolve varCount allClasses cs = runIdentity $ runExceptT $ flip runReaderT allClasses $ solver varCount st
   where
     st = (emptySubst, cs)
 
-unifyMany :: [TypeError SourcePos] -> [InfernoType] -> [InfernoType] -> Solve Subst
+unifyMany :: [TypeError SourcePos] -> [InfernoType] -> [InfernoType] -> SolveState Int Subst
```
siddharth-krishna, line 1265: Once everything works, I can clean this up and use some `FreshVarState` or something instead of `Int`

inferno-core/src/Inferno/Infer.hs
```diff
@@ -1399,9 +1563,11 @@ findTypeClassWitnesses allClasses iters tyCls tvs =
             ((Subst $ Map.fromList $ map snd found) :) <$> getSolutions ((\x -> x - 1) <$> i)
           _ -> pure []
 
+-- NOTE: this function resets the fresh variable counter each time it is called
+-- TODO confirm this doesn't matter
```
siddharth-krishna, line 1567: @goodlyrottenapple should be fine, right?

inferno-core/src/Inferno/Instances/Arbitrary.hs
```diff
@@ -227,7 +232,7 @@ deriving instance ToADTArbitrary Fixity
 
 deriving via (GenericArbitrary Fixity) instance Arbitrary Fixity
 
-instance Arbitrary pos => Arbitrary (Comment pos) where
+instance (Arbitrary pos) => Arbitrary (Comment pos) where
```
siddharth-krishna, line 235: Sorry the rest of the changes in this file are formatting changes like this one... think my ormulu version changed or something

inferno-core/src/Inferno/Parse.hs
```diff
@@ -362,6 +362,36 @@ array p = label "array\nfor example: [1,2,3,4,5]" $
           )
         <|> pure []
 
+record :: Parser a -> Parser (SourcePos, [(Ident, a, Maybe SourcePos)], SourcePos)
+record p = label "record\nfor example: {name: \"Zaphod\", age: 391}" $
```
siddharth-krishna, line 366: Still haven't changed the parser to parse records as `{x = 1.1, y = 2.2}` and record field access as `rec.field`

inferno-types/src/Inferno/Types/Syntax.hs
```diff
@@ -1435,23 +1519,30 @@ prettyPrec isBracketed prec expr =
               <> elsePretty
     Op e1 _ _ (n, NoFix) ns (Ident op) e2 ->
       bracketWhen e2 (prec > n) $
-        prettyOpAux (n + 1) e1 <> (if hasTrailingComment e1 then hardline else mempty)
+        prettyOpAux (n + 1) e1
+          <> (if hasTrailingComment e1 then hardline else mempty)
```
siddharth-krishna, line 1523: Most of these are also formatting changes...

inferno-types/src/Inferno/Types/VersionControl.hs
```diff
@@ -108,13 +109,13 @@ class VCHashUpdate obj where
   default (&<) :: (Generic obj, GenericVCHashUpdate (Rep obj)) => Context SHA256 -> obj -> Context SHA256
   ctxt &< o = genHashUpdate ctxt $ from o
 
-hashUpdateVia :: ByteArrayAccess ba => (obj -> ba) -> Context SHA256 -> obj -> Context SHA256
+hashUpdateVia :: (ByteArrayAccess ba) => (obj -> ba) -> Context SHA256 -> obj -> Context SHA256
```
siddharth-krishna, line 112: This file is also mostly formatting...

siddharth-krishna (commented):
@goodlyrottenapple if you ever have a few moments this week, I'd love to get your thoughts on the changes to the type inference. I've mentioned you on some comments with some particular questions. Thanks!

inferno-core/src/Inferno/Eval.hs
```diff
@@ -204,6 +193,16 @@ eval env@(localEnv, pinnedEnv) expr = case expr of
       _ -> throwM $ RuntimeError "failed to match with a bool"
   Tuple_ es ->
     foldrM (\(e, _) vs -> eval env e >>= return . (: vs)) [] (tListToList es) >>= return . VTuple
+  Record_ fs -> do
+    valMap <- foldrM (\(f, e, _) vs -> eval env e >>= \v -> return ((f, v) : vs)) [] fs
+    return $ VRecord $ Map.fromList valMap
+  RecordField_ (Ident r) f -> do
+    eval env (Var undefined Nothing LocalScope $ Expl $ ExtIdent $ Right r) >>= \case
```
Daniel-Diaz, line 200: I guess this `undefined` is safe, but maybe a short comment explaining why?

inferno-types/src/Inferno/Types/Value.hs
```diff
@@ -42,6 +44,7 @@ data Value custom m
   | VEnum VCObjectHash Ident
   | VArray [Value custom m]
   | VTuple [Value custom m]
+  | VRecord (Map.Map Ident (Value custom m))
```
Daniel-Diaz, line 47: Cool. It reminds me of the evaluation environment!

Daniel-Diaz (approved):

inferno-core/src/Inferno/Infer.hs
goodlyrottenapple: Yes, i think so

inferno-core/src/Inferno/Eval.hs
albertov: An `error` with a searchable message also works and if the impossible happens one has something to search.

goodlyrottenapple: Could alternatively just inline the eval call rhs:

```haskell
case Map.lookup (ExtIdent $ Right r) localEnv of
      Just v -> return v
Nothing -> throwM $ RuntimeError $ show (ExtIdent $ Right r) <> " not found in the unpinned env"
```

siddharth-krishna:
@Daniel-Diaz I found some bugs in the PR, so I'm pushing a few new commits with some extra tests and the respective bug fixes. Hope you don't mind reviewing the new commits in the same PR so I can merge it in in a good state. :)

The changes are:
- Add the new record constructors of `Expr` and `InfernoType` to the `Arbitrary` instances, so that they are covered by the QuickCheck tests
- Fix the pretty printing of records to pass the pretty-printer-and-parser-are-inverses test
- Fix the fresh variable counter in `inferTypeReps` as suggested by Sam
- Add tests for the above

I also updated the PR description with details of the new record syntax and types.

Daniel-Diaz (approved):
:shipit:

PR #105, opened 2024-03-18
HLint everything and add to CI

siddharth-krishna:
Spring cleaning: thought I'd get rid of all the blue squiggles while there weren't many open PRs. :)

I ran:
```bash
cabal install apply-refact hlint
for f in `git ls-files | grep -e '.*\.hs'`; do echo linting $f; hlint --refactor --refactor-options="-i" $f; done
```
And manually checked that nothing semantically changed.

~Also trying out a GitHub action that should generate lint suggestions in the PRs.~
I thought the action would add PR suggestions in the web interface which could be applied with one click, but it didn't do that, and going through the CI logs feels a bit painful. This other action might do that, but it needs some security/token set up so I'll save it for another time:
https://github.com/haskell-actions/hlint-scan

For now, it just adds a hint to the PR diff view:
image

### DIFF
--- a/.github/hlint.json
+++ b/.github/hlint.json
@@ -1,0 +1,19 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "hlint",
+      "pattern": [
+        {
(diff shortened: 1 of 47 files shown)


albertov (approved):
🧼

PR #109, opened 2024-03-28
Move `inferno-ml-server` to this repo

ngua:
Moves most of the things related to `inferno-ml-server` to this repo, including:
- the `inferno-ml-server` Haskell package 
- the base NixOS modules and image configurations
- the server `nixosTest`

There are no changes to `inferno-ml-server` itself beyond some minor renaming in some places, updated comments, etc...

To make it easier to extend the image configurations, I've also added an `image-config` flake input. This should contain additional NixOS modules to include in the image configuration (they will be automatically included). For this repo, we don't need any additional config, so it's set to a new `nix/dummy` directory. But it can be overridden elsewhere. I've tried this out with a different repo and it works nicely

I've also included some of the Python packages that I made for training. These aren't strictly related to `inferno-ml-server` but they might be useful to have here (we also have `devShells..pytorch` already, so it's in the same vein as that). If we really don't want them we can put them somewhere else

I also noticed that the flake was getting a bit out of control, so I split everything into smaller modules. There's a comment in the flake explaining the new structure. This does not affect actually using the flake -- it's just for internal organization to make things nicer and more maintainable/readable


siddharth-krishna (commented):
Looks good, but I notice that some test files seem to be duplicated, e.g. `nix/inferno-ml/tests/models/mnist.ts.pt` is also at `inferno-ml/test/mnist.ts.pt` -- is it possible to de-duplicate?

ngua:
@siddharth-krishna Oh, good catch, I guess those are from renaming/moving some things. I'll remove the redundant ones

siddharth-krishna:
Thanks, but I notice we still have two `mnist.ts.pt` files.. I was hoping to keep the repo size down

ngua:
@siddharth-krishna I only see one at `nix/inferno-ml/tests/models/mnist.ts.pt`, where is the other one?

siddharth-krishna:
My bad, I must not have read the find-grep output properly!

PR #113, opened 2024-04-03
[inferno-ml-server] Add `terminated` to DB types

ngua:
This adds a `terminated` column to most of the DB tables, corresponding to a `Maybe UTCTime` in the Haskell types. This lets us track when DB types have been "deleted":

- a `Just time` means that the DB row was "deleted" at `time`; `inferno-ml-server` will not use these rows for inference evaluation or model caching
- `Nothing` means that the field is still active

Originally I was going to use a Postgres enum of `active`/`passive`. But then we would need another column to record the time, which would be messy and redundant, since the semantics of a nullable `timestamptz` are essentially the same


inferno-ml-server/src/Inferno/ML/Server/Inference.hs
```diff
@@ -247,7 +247,12 @@ getParameter iid =
     =<< queryStore q (Only iid)
   where
     q :: Query
-    q = [sql| SELECT * FROM params WHERE id = ? LIMIT 1 |]
+    q =
+      [sql|
+        SELECT * FROM params
+        WHERE id = ? AND terminated IS NULL
+        LIMIT 1
+      |]
```
albertov, line 255: I can't help mentioning https://hackage.haskell.org/package/postgresql-typed when I see quasi-quoted SQL intended for postgres :) It's the haskell equivalent of other "infer SQL types at compile time" libraries like https://ocaml.org/p/pgocaml_ppx/ or https://github.com/launchbadge/sqlx

albertov (approved):
Looks good and a nice way to model it

PR #119, opened 2024-05-02
ML: Add write type and have scripts return writes instead of calling bridge

siddharth-krishna:
The tests work now, but we can hold off on merging this until we've checked that these changes are compatible with the orchestrator.


ngua (approved):
This seems to be working well, good performance too, so I think we can merge this now. We can potentially revisit the default chunk size in the output stream later, but at the moment it seems good to me

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -703,6 +704,7 @@ data IValue
   | ITuple (IValue, IValue)
   | ITime EpochTime
   | IEmpty
+  | IArray (Vector IValue)
```
siddharth-krishna, line 707: Is `IValue` the type of all values that can be stored in input/output PIDs in TachDB? I don't think they support arrays, if so.

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -715,10 +717,19 @@ instance FromJSON IValue where
...
       | [x, y] <- Vector.toList a ->
-          fmap ITuple $
-            (,) <$> parseJSON x <*> parseJSON y
-      | Vector.null a -> pure IEmpty
-    _ -> fail "Expected one of: string, double, time, tuple, unit (empty array)"
+          (,) <$> parseJSON x <*> parseJSON y <&> \case
+            -- We don't want to confuse a two-element array of tuples with
+            -- a tuple itself. For example, `"[[10.0, {\"time\": 10}], [10.0, {\"time\": 10}]]"`
+            -- should parse as a two-element array of `(double, time)` tuples,
+            -- not as a `((double, time), (double, time))`. I can't think of
```
siddharth-krishna, line 724: Haha yeah I see the issue, and agree with your solution, but btw I think inferno allows n-tuples for any n != 1. As with my previous comment, I don't think Tach supports tuples, so they're only script-internal values and we wouldn't need to support them in the interface to Tach.

siddharth-krishna (commented):
Thanks Rory, I had put a bunch of hacks in to get the integration tests running quickly, and your changes make it all nicer and neater. :)

PR #121, opened 2024-05-22
[inferno-ml] Save evaluation info

ngua:
Adds support for getting total elapsed time, CPU time, and memory allocations every time an inference param is run in `inferno-ml-server`

Originally I thought we could use EKG for this, but then we would need to keep `inferno-ml-server` on to get the evaluation info. Given our constraints elsewhere, we should avoid this. So I've continued my quest to Use Postgres For Everything™ and stored it in the DB

### DIFF
--- a/inferno-ml-server/src/Inferno/ML/Server/Inference.hs
+++ b/inferno-ml-server/src/Inferno/ML/Server/Inference.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE LexicalNegation #-}
+{-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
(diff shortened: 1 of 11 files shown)


inferno-ml-server/src/Inferno/ML/Server/Inference.hs
```diff
@@ -235,7 +252,67 @@ runInferenceParam ipid (fromMaybe 128 -> res) =
...
-        . (. (* 1000000) . fromIntegral)
+        . (. (* 1_000_000) . fromIntegral)
+
+    withEvaluationInfo :: RemoteM a -> RemoteM a
+    withEvaluationInfo f = withRunInIO $ \r -> do
+      -- So allocation counter doesn't go below the lower limit, which is
+      -- unlikely but should be accounted for at any rate
+      setAllocationCounter maxBound
+      start <- getCurrentTime
+      bytes0 <- getAllocationCounter
```
ngua, line 263: An alternative to this would be to get one of the memory-related fields from `GHC.Stats.RTSStats` and compare it. But I noticed this was very unreliable unless a GC was triggered before and after -- sometimes the difference between beginning/ending max memory in use was negative. Of course we could trigger a GC before and after eval, but that can cause performance to degrade. So I think getting allocated bytes is sufficient

mirko-plowtech (approved):
LGTM!

PR #122, opened 2024-05-30
[inferno-ml] Change input/output representation

ngua:
Changes the internal representation of `InferenceParam` inputs/outputs. Instead of having two separate `inputs` and `outputs` as vectors, they are now both in a single field and annotated with a `ScriptInputType`, which controls readability/writability of the input. Also, instead of being a `Vector`, they are now explicitly mapped to Inferno `Ident`s

The rationale for the merging `inputs`/`outputs`:
- We need to provide outputs as well as inputs as arguments during script evaluation. If we don't do this, the identifiers used in `makeWrites` will not resolve to anything
- We can't just concatenate the old `inputs` and `outputs` fields, because users may wish to have identically named inputs and outputs; for example, they may want to both read from and write to `input0`. This would lead to the following during script eval:
  ```
  fun input0 input0 -> ...
  ``` 
- Accordingly, we need some way of allowing identically named identifiers to be both inputs and outputs. Combining the two argument types and storing if they are readable, writable, or both solves this

The rationale for storing the Inferno identifiers:
- When scripts are created (e.g. via an Inferno LSP server), the only thing that is provided is the list of Inferno `Ident`s
- These `[Ident]` need to correspond exactly to the `inputs` that the `InferenceParam` contains
- Even if we sort this list of identifiers, we need to make sure that the order of the `InferenceParam`'s inputs is exactly the same as the order of the original `[Ident]`. This could lead to bug-prone assumptions or workarounds, since we would't have the original `[Ident]` in the param
- To make sure that we always know which order is correct, we can store the `Ident`s along with the actual inputs in a `Map` and then use `Map.toAscList` to get the correct order for script arguments

Also I fixed the JSON encoding/decoding so that NaNs can be transmitted to/from `inferno-ml-server`. Previously, `[null, 0.0, 0.0]` would parse to `[IEmpty, IDouble 0.0, IDouble 0.0]`, which is of course incorrect. Now it correctly parses to `[IDouble NaN, IDouble 0.0, IDouble 0.0]`


ngua:
CI is broken due to IOG cache woes (seems to be a Cloudflare issue)

albertov (approved):
looks good

PR #124, opened 2024-06-07
[inferno-ml] Link multiple models to params

ngua:
Adds the ability to link multiple `ModelVersion` IDs to inference params and updates `inferno-ml-server` script evaluator to copy all models linked to a given param

Also adds configurable log level for `inferno-ml-server`

### DIFF
--- a/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
+++ b/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
@@ -600,9 +600,16 @@ data InferenceParam uid gid p s = InferenceParam
     -- For existing inference params, this is the foreign key for the specific
     -- script in the 'InferenceScript' table (i.e. a @VCObjectHash@)
     script :: s,
-    -- | This needs to be linked to a specific version of a model rather
-    -- than the @model@ table itself
-    model :: Id (ModelVersion uid gid Oid),
+    -- | All of the (specific versions of) models that can be used with this
+    -- parameters. @inferno-ml-server@ will copy the contents of each of the
+    -- model versions when evaluating inference scripts. Inference scripts can
+    -- reference any of the linked model versions by referring to the parent
+    -- model\'s name, e.g. @loadModel "name.ts.pt"@
+    --
+    -- Each element represents the ID of a specific model version. However, due
+    -- to limitations in PostgreSQL, there is no referential integrity; i.e.
+    -- the elements are treated as plain integers
+    models :: Vector (Id (ModelVersion uid gid Oid)),
     -- | This is called @inputs@ but is also used for script outputs as
     -- well. The access (input or output) is controlled by the 'ScriptInputType'.
     -- For example, if this field is set to @[("input0", Single (p, Readable))]@,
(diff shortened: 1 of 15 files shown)


inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -600,9 +600,16 @@ data InferenceParam uid gid p s = InferenceParam
...
-    model :: Id (ModelVersion uid gid Oid),
+    -- | All of the (specific versions of) models that can be used with this
+    -- parameters. @inferno-ml-server@ will copy the contents of each of the
+    -- model versions when evaluating inference scripts. Inference scripts can
+    -- reference any of the linked model versions by referring to the parent
+    -- model\'s name, e.g. @loadModel "name.ts.pt"@
+    --
+    -- Each element represents the ID of a specific model version. However, due
+    -- to limitations in PostgreSQL, there is no referential integrity; i.e.
+    -- the elements are treated as plain integers
```
albertov, line 611: What PostgreSQL limitations prevent referential integrity? AFAICT this looks like a many-to-one relation between modelVersion and inferenceParam which can be modeled as a inferenceParam foreign key on the modelVersion table

nix/inferno-ml/migrations/v1-create-tables.sql
```diff
@@ -66,7 +66,11 @@ create table if not exists params
   ( id serial primary key
     -- Script hash from `inferno-vc`
   , script bytea not null references scripts (id)
-  , model integer references mversions (id)
+    -- Array of model version IDs; unfortunately we lose referential
+    -- integrity because Postgres does not allow arrays of foreign keys.
+    -- However, we don't allow model version deletion anyway, which
+    -- mitigates this to some degree
+  , models integer[] not null
```
albertov, line 73: I see now why. So this is a many-to-many relationship, right? To regain referential integrity this could be modeled as
```sql
create table model_param (
  model integer references mversions (id) on update cascade on delete cascade,
  param integer references params (id) on update cascade on delete cascade
)
```
And queries adjusted accordingly using `array_agg` to flatten the left join between `param` and `model_param` back into an array

nix/inferno-ml/migrations/v1-create-tables.sql
```diff
@@ -92,5 +96,32 @@ create table if not exists evalinfo
...
+        )
+    then
+      raise exception 'ID % is not a valid model version primary key', mv;
+    end if;
+  end loop;
+  return new;
+end;
+$$ language plpgsql;
+
+create trigger "verify-param-models" before insert or update on params
```
albertov, line 123: if going this route, a trigger is also needed to prevent modelversions which are referrenced by params from being deleted

albertov (approved):
Looks good but I believe referential integrity doesn't need to be sacrificed, I left a comment explaining how

ngua:
Thanks @albertov, I did consider using a junction table as you suggested. However, we don't allow deletion of anything at the moment (including params, models, model versions), so I thought the extra indirection might not be worth it and the trigger would suffice. And my biggest concern with the trigger would be performance degradation, but I don't think we'll have a particularly heavy load with param insertions/updates. I think I'll keep it as-is for now but we can revisit later in case needs change

ngua:
Or I might change it in a follow-up PR, since it's the more correct way to do it at any rate. However, I have another in-progress Inferno ML thing that I'd like to finish first

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
ngua: The reason I phrased it like that is that arrays of foreign keys are in a recent SQL standard but Postgres doesn't support them

PR #125, opened 2024-06-19
[inferno-ml] Move `BridgeInfo` to DB, add more instances

ngua:
Various changes:
- Moves `BridgeInfo` to DB instead of having convoluted endpoint and local caching
- Updates script evaluator accordingly
- Adds some useful instances for different types (e.g. `InferenceParam`) so we don't need orphans elsewhere anymore

### DIFF
--- a/inferno-ml-server/src/Inferno/ML/Server/Bridge.hs
+++ b/inferno-ml-server/src/Inferno/ML/Server/Bridge.hs
@@ -1,8 +1,7 @@
-{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE QuasiQuotes #-}
 
 module Inferno.ML.Server.Bridge
-  ( registerBridgeInfo,
-    getBridgeInfo,
+  ( initializeInferno,
   )
 where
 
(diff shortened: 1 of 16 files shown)


inferno-ml-server/src/Inferno/ML/Server/Bridge.hs
```diff
@@ -29,50 +29,45 @@ import Servant.Client.Streaming
...
-    funs = mkBridgeFuns valueAt latestValueAndTimeBefore valuesBetween
-
-    valueAt :: Int64 -> PID -> EpochTime -> RemoteM IValue
-    valueAt res pid = callBridge . Bridge.valueAtC res pid
+    -- There should always be a bridge saved for the param
+    getBridgeInfo :: RemoteM BridgeInfo
+    getBridgeInfo = firstOrThrow NoBridgeSaved =<< queryStore q (Only ipid)
+      where
+        q :: Query
+        q = [sql| SELECT * FROM bridges WHERE id = ? |]
```
mirko-plowtech, line 47: ohh this syntax `[sql| |]` would be really nice on sqlite, I'm wondering if there is some lib out there.

mirko-plowtech (approved):
LGTM!

inferno-ml-server/src/Inferno/ML/Server/Bridge.hs
ngua: I think there's a raw string quasi-quoter that works with `sqlite-simple`: https://hackage.haskell.org/package/raw-strings-qq

PR #130, opened 2024-07-18
[inferno-vc] Modify cached client so fetches for remote objects are serialized

albertov:
This avoids concurrent requests for the same object which is causing the thundering herd problem when several requests for the same object are DOSing inferno-vc


siddharth-krishna (approved):
Nice one!

siddharth-krishna:
Ah sorry, would you mind bumping version numbers and adding to the changelog? We don't have plow-lint on this public repo to force us to do it, and this seems like an important change.

PR #137, opened 2024-10-11
[inferno-vc] Made fetchObjectClosureHashes return the root scriptId and added logging to cached client and server

albertov:
Made fetchObjectClosureHashes return the root scriptId because it belongs to the closure and so it is coherent with fetchObjectClosure. This makes the cached client take a lock on the requested scriptHash too when fetching the object which it wasn't doing before.

Also added logging to cached client to see hits/missed and to server to see which scriptIds are being used to call fetchObjects and fetchObjectClosureHashes


mirko-plowtech (approved):
LGTM!

PR #138, opened 2024-10-11
Revert "[inferno-vc] Made fetchObjectClosureHashes return the root scriptId and added logging to cached client and server"

brent80:
Reverts plow-technologies/inferno#137


smurphy8:
@albertov we are reverting this so it doesn't accidentally sneak back in till you get a chance to check it out.

smurphy8 (approved):

PR #147, opened 2024-10-31
[inferno-ml] Change parameter input/output representation

ngua:
The previous way that parameter inputs/outputs were stored, with both condensed into a single `Map Ident ...`, turned out to be quite nightmarish to work with. We are going to be doing things like filtering or querying parameters by output, so we need it to be easier to look outputs up. So this changes the inputs/outputs back to the way they used to be, e.g. two separate `Map`s


Daniel-Diaz (approved):
~~This is a bit different, as it now allows an identifier which has input and no output (or viceversa), which wasn't possible with the single-map approach. I don't know the code enough to know whether this is an advantage or a disadvantage. Just wanted to point it out.~~

EDIT: Disregard the above comment. It was a misunderstanding.

PR #149, opened 2024-11-29
[inferno-ml] Use connection pool

ngua:
Currently, the entire server runs within `withConnect` and holds a single `Connection` within its `Env`. If Postgres is unreachable for some reason when the server starts, the process will hang until it's killed by an exception (which takes quite some time). If the connection fails while the server is running, the entire process will be brought down. This switches the server to use a `Pool` from `resource-pool` instead to create/hold connections on demand

### DIFF
--- a/inferno-ml-server/src/Inferno/ML/Server.hs
+++ b/inferno-ml-server/src/Inferno/ML/Server.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE OverloadedRecordDot #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module Inferno.ML.Server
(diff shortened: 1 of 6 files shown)


inferno-ml-server/src/Inferno/ML/Server.hs
```diff
@@ -72,19 +72,19 @@ main = runServer =<< mkOptions
...
-  withConnect (view #store cfg) $ \conn ->
-    f
-      =<< Env cfg conn tracer
-        <$> newMVar ()
-        <*> newEmptyMVar
-        <*> newManager defaultManagerSettings
-        <*> newIORef Nothing
+  f
+    =<< Env cfg tracer
+      <$> newConnectionPool cfg.store
```
albertov, line 83: It might be reasonable to destroy the pool when done so it closes connections explicitly. With sqlite, failing to do so left dangling WAL files which are annoying. With postgres it may leave open resources in the db server until a timeout expires (not sure but it wouldn’t surprise me)

albertov (approved):
Lgtm but left comment

inferno-ml-server/src/Inferno/ML/Server.hs
ngua: Good idea, I've added that

PR #155, opened 2025-01-21
Make VCHashUpdate @CTime always encode as 64 bits

albertov:
This makes vcHash for VCMeta produce the same hash on 32bit and 64 bit plarforms


albertov:
Supersedes #150

smurphy8 (approved):

PR #157, opened 2025-01-23
[nix] Drop GHC 8 support

ngua:
It's been quite a while since we've actually needed GHC 8 support in `inferno`, so I'm dropping all references to it from the build and the hacky workaround to get GHC 8 and 9 working at the same time


mirko-plowtech (approved):
:rocket:

PR #158, opened 2025-01-28
[inferno-ml] Remove existing images, replace with NixOS modules

ngua:
This removes the pre-defined Inferno ML images from `legacyPackages` and adds various image-related `nixosModules` to the flake outputs. This way we can use the NixOS modules defined in this repo more flexibly elsewhere. This also removes some hacky things used to override modules (most importantly the `image-config` flake input)


ngua:
This is ready to merge once we take care of some things elsewhere. But I've confirmed that this approach works (and works nicely) by using the new `nixosModules` to build things

albertov (approved):

PR #159, opened 2025-02-12
Upgrade to GHC 9.10, bump `nixpkgs`

ngua:
- Updates `haskell-nix` input to recent commit
- Upgrades nixpkgs accordingly
- Changes compiler version to GHC 9.10
- Changes formatter to Fourmolu (see notes below for more)
- Upgrades NVIDIA drivers for V100 GPU (amazingly working somehow)
- Fixes upgrade-related problems with VSCode packages for Inferno
- Fixes some overlay and haskell.nix related configuration for newer version of Hasktorch

Everything is building fine, including our CPU/GPU Inferno ML images (tested elsewhere, running on actual EC2 instances). I didn't need to make any changes to the Haskell code for the upgrade, just configuration changes.

**NOTE**: There are some compiler warnings now that are present on GHC 9.10 but not on earlier versions. I'd prefer not to make a CPP mess of a large number of Haskell modules, so I've left those unaddressed. Once we switch to GHC 9.10+ everywhere, we can fix them.


flake.nix
```diff
@@ -188,31 +201,26 @@
...
+            programs = {
+              nixpkgs-fmt.enable = true;
+              ormolu = {
+                enable = true;
+                package =
+                  let
+                    # Using `hackage-package` will prevent building `ormolu`
+                    # from interfering with the build plan (incl. incompatible
+                    # compiler versions)
+                    o = pkgs.haskell-nix.hackage-package {
```
ngua, line 213: I've switched the formatter to Fourmolu, for several reasons:
- We use it elsewhere
- It's better than Ormolu
- Even staying with Ormolu will require a large reformatting of the Haskell codebase, so we might as well switch to Fourmolu

I've disabled the CI check for formatting for now. Once the upgrade is in and confirmed working everywhere, I will re-enable it in a re-formatting PR

inferno-core/inferno-core.cabal
```diff
@@ -1,107 +1,114 @@
...
-category:            DSL,Scripting
-homepage:            https://github.com/plow-technologies/inferno.git#readme
-bug-reports:         https://github.com/plow-technologies/inferno.git/issues
-copyright:           Plow-Technologies LLC
-license:             MIT
-license-file:        LICENSE
-author:              Sam Balco
-maintainer:          info@plowtech.net
-build-type:          Simple
+cabal-version:      2.4
```
ngua, line 1: I `cabal-fmt`'d the Cabal files, the inconsistency was bugging me. I can add a formatting check for this in the future re-formatting PR

inferno-core/inferno-core.cabal
```diff
@@ -1,107 +1,114 @@
...
+    -Wall -Wunused-packages -Wincomplete-uni-patterns
+    -Wincomplete-record-updates -Wincomplete-patterns
+
   build-depends:
-      base                      >= 4.13 && < 4.17
-    , bimap                     >= 0.5.0 && < 0.6
-    , bytestring                >= 0.10.10 && < 0.12
-    , containers                >= 0.6.2 && < 0.7
-    , cryptonite                >= 0.30 && < 0.31
+    , base
```
ngua, line 58: I decided to get rid of the strict version bounds, again for several reasons:
- We use Nix for this and don't need to involve the Cabal bounds solver again
- This is our practice elsewhere
- We added these version bounds only to appease `cabal sdist` for release to Hackage, but we haven't done a Hackage release in ~2 years. I don't think we even should, since a lot of the project would be very difficult to build outside of Nix
- Having these constraints makes upgrading harder
- `cabal gen-bounds` doesn't help very much

inferno-ml/app/Main.hs
```diff
@@ -14,13 +14,15 @@ import System.Environment (getArgs)
...
 main = do
-  file <- head <$> getArgs
-  src <- Text.readFile file
-  Interpreter {evalExpr, defaultEnv, parseAndInferTypeReps} <-
-    mkInferno @_ @(MlValue ()) mlPrelude customTypes
-  case parseAndInferTypeReps src of
-    Left err -> print err
-    Right ast ->
-      evalExpr defaultEnv Map.empty ast >>= \case
+  getArgs >>= \case
```
ngua, line 17: I tested this executable fairly early on so changed it to avoid `head` (which emits a warning on GHC 9.10). However, there are unfortunately several other places in the codebase where `head` is used. I've decided to leave that for another day, as it involves a fairly large refactor (either that or disabling the partial warning, which I think is a good thing and should be left enabled)

nix/default.nix
```diff
@@ -1,19 +1,17 @@
-{ compiler ? "ghc925"
+{ compiler ? "ghc910"
 , config
 , ghcOptions ? [ ]
 , profiling ? false
-  # Must be of the form: { device = <cpu|cuda-10|cuda-11>; }
+  # Must be of the form: { device = <cpu|cuda-118>; }
```
ngua, line 5: We can only build with CUDA 11.8 now

nix/overlays/nvidia/v100.nix
```diff
@@ -1,17 +1,17 @@
 final: prev: {
   # Override the default NVIDIA driver with one known to work well with
-  # CUDA
+  # the V100 GPU
```
ngua, line 3: This has been tested and is working on a V100 GPU

nix/overlays/torch.nix
```diff
@@ -1,20 +1,49 @@
...
+    # We only care about supporting this platform for now. Darwin support
+    # isn't necessary
+      assert system == "x86_64-linux";
+      assert lib.assertOneOf "device" device [ "cpu" "cuda-118" ];
       assert device == "cpu" -> !cudaSupport;
-      final.callPackage "${final.hasktorch}/nix/libtorch.nix" {
-        inherit device cudaSupport;
-      };
+
+      (prev.libtorch-bin.override { inherit cudaSupport; }).overrideAttrs
```
ngua, line 14: Hasktorch is on a slightly older version of `libtorch-bin` and Pytorch (2.3.0), hence the overrides

nix/modules/ml.nix
```diff
@@ -27,45 +28,46 @@
     };
   };
 
-  nonReinstallablePkgs = [
```
ngua, line 30: IIRC I originally added these only so that Hasktorch would build, back when we first began the Hasktorch integration. However, it seems that it no longer builds with `nonReinstallablePkgs` specified. Removing it entirely works

albertov (approved):

PR #165, opened 2025-02-28
Moved RemoteTrace to inferno-ml-server-types

albertov:
And made it JSON de/serializable. In preparation for persistent traces.

### DIFF
--- a/inferno-ml-server-types/inferno-ml-server-types.cabal
+++ b/inferno-ml-server-types/inferno-ml-server-types.cabal
@@ -36,6 +36,7 @@ library
     Inferno.ML.Server.Client
     Inferno.ML.Server.Client.Bridge
     Inferno.ML.Server.Types
+    Inferno.ML.Types.Value
 
   hs-source-dirs:  src
   build-depends:
@@ -68,3 +69,7 @@ library
     , unix
     , uuid
     , vector
+    , bson
+    , hasktorch
+    , prettyprinter
+    , template-haskell
--- a/inferno-ml-server-types/inferno-ml-server-types.cabal
+++ b/inferno-ml-server-types/inferno-ml-server-types.cabal
@@ -1,6 +1,6 @@
 cabal-version: 2.4
 name:          inferno-ml-server-types
-version:       0.12.2
+version:       0.13.0
 synopsis:      Types for Inferno ML server
 description:   Types for Inferno ML server
 homepage:      https://github.com/plow-technologies/inferno.git#readme
--- a/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
+++ b/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
@@ -1065,13 +1065,17 @@ data RemoteError
   | InvalidScript Text
   | InvalidOutput Text
   | -- | Any error condition returned by Inferno script evaluation
-    InfernoError String
+    InfernoError SomeInfernoError
   | NoBridgeSaved
   | ScriptTimeout Int
   | ClientError String
   | OtherRemoteError Text
   deriving stock (Show, Eq, Generic)
   deriving anyclass (ToJSON, FromJSON)
+
+newtype SomeInfernoError = SomeInfernoError String
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (Exception, ToJSON, FromJSON)
 
 instance Exception RemoteError where
   displayException = \case
@@ -1102,7 +1106,7 @@ instance Exception RemoteError where
         [ "Script output should be an array of `write` but was"
         , Text.unpack t
         ]
-    InfernoError x ->
+    InfernoError (SomeInfernoError x) ->
       unwords
         [ "Inferno evaluation failed with:"
         , x
--- a/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
+++ b/inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
@@ -7,9 +7,11 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE OverloadedRecordDot #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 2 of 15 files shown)


nix/default.nix
```diff
@@ -63,6 +63,7 @@ pkgs.haskell-nix.cabalProject {
       config.treefmt.build.wrapper
       pkgs.torch
       pkgs.torch.dev
+      pkgs.ghcid
```
albertov, line 66: Does not add it to shell tools because https://github.com/haskell-fswatch/hinotify/pull/39

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -1043,6 +1055,212 @@ data EvaluationEnv gid p = EvaluationEnv
...
+    NoSuchModel (Either (Id (Model (EntityId GId))) (Id (ModelVersion (EntityId GId) Oid)))
+  | NoSuchScript VCObjectHash
+  | NoSuchParameter (Id (InferenceParam (EntityId GId) PID))
+  | InvalidScript Text
+  | InvalidOutput Text
+  | -- | Any error condition returned by Inferno script evaluation
+    InfernoError SomeInfernoError
+  | NoBridgeSaved
+  | ScriptTimeout Int
+  | ClientError String
```
albertov, line 1071: Changed this to `String` because `ClientError` is not de/serializable

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -1043,6 +1055,212 @@ data EvaluationEnv gid p = EvaluationEnv
...
+  | OtherRemoteError Text
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (ToJSON, FromJSON)
+
+newtype SomeInfernoError = SomeInfernoError String
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (Exception, ToJSON, FromJSON)
```
albertov, line 1078: Changed from existential over `Show` because existential is not trivially deserializable and not really needed here

inferno-ml-server-types/inferno-ml-server-types.cabal
```diff
@@ -36,33 +36,38 @@ library
...
     , base
+    , bson
     , bytestring
     , conduit
     , containers
     , deepseq
     , extra
     , generic-lens
     , hashable
+    , hasktorch
```
ngua, line 54: Unfortunately we can't use `hasktorch` as a dependency without dragging the whole thing into `all`. We're on incompatible GHC versions, and `inferno-ml-server-types` is a dependency in `all`

ngua (changes requested):

ngua:
@albertov If there's something that depends on `hasktorch` that we need here, we will need to change it to something else and do a conversion elsewhere. Hasktorch uses GHC libraries that I think will be incompatible between the version of GHC in `inferno` and the one in `all`

Also, `hasktorch` brings in a large amount of non-Haskell dependencies that will complicate things elsewhere

Either way, we would either need to figure out how to include `hasktorch` in `all`, or change things here so that there's no `hasktorch` dependency

albertov:
> @albertov If there's something that depends on `hasktorch` that we need here, we will need to change it to something else and do a conversion elsewhere. Hasktorch uses GHC libraries that I think will be incompatible between the version of GHC in `inferno` and the one in `all`
> 
> Also, `hasktorch` brings in a large amount of non-Haskell dependencies that will complicate things elsewhere
> 
> Either way, we would either need to figure out how to include `hasktorch` in `all`, or change things here so that there's no `hasktorch` dependency

hasktorch is needed because MlValue but MlValue is only used to define instances for `MlValue BridgeValue` afaict. I'll try moving MlValue back to where it was leave the instances as orphans see if it's possible

inferno-ml-server-types/src/Inferno/ML/Server/Types.hs
```diff
@@ -1043,6 +1055,212 @@ data EvaluationEnv gid p = EvaluationEnv
 instance (Arbitrary p) => Arbitrary (EvaluationEnv gid p) where
   arbitrary = genericArbitrary
 
+data RemoteError
```
ngua, line 1058: Things like `EntityId` and `PID` are to make things compatible with `all`, where we have similar types. But we can't use those types in `inferno` since it's an OSS project

In the Inferno ML server API, things need to be instantiated with concrete types. For example:

```
type InferenceAPI gid p =
  "inference"
    :> "run"
    :> Capture "id" (Id (InferenceParam gid p))
    :> QueryParam "res" Int64
    :> QueryParam' '[Required] "uuid" UUID
    :> StreamPost NewlineFraming JSON (WriteStream IO)
```

The `gid` and `p` are instantiated differently between `inferno` and `all`. If we just use the `EntityId` and `PID` types that are currently defined in `inferno-ml-server` (not the types package), there will be a conflict between them when imported in `all` (because there it's a different `PID`)

I know this is complicated, but it's because of the split between the two projects. Merging them into a single project would be a lot of work and in the past was not possible

ngua:
@albertov There are integration tests for IMLS, `nix build -L .#checks.x86_64-linux.inferno-ml-server`, I'm not sure if that's documented. Could you run those manually in case you haven't yet? They don't run on CI because it might take too long on GC Actions

albertov:
> @albertov There are integration tests for IMLS, `nix build -L .#checks.x86_64-linux.inferno-ml-server`, I'm not sure if that's documented. Could you run those manually in case you haven't yet? They don't run on CI because it might take too long on GC Actions

```console
....
vm-test-run-inferno-ml-server-test> (finished: run the VM test script, in 15.33 seconds)
vm-test-run-inferno-ml-server-test> test script finished in 15.35s
vm-test-run-inferno-ml-server-test> cleanup
vm-test-run-inferno-ml-server-test> kill machine (pid 9)
vm-test-run-inferno-ml-server-test> qemu-system-x86_64: terminating on signal 15 from pid 6 (/nix/store/c9m6yd8fg1flz2j5r4bif1ib5j20a0cy-python3-3.12.8/bin/python3.12)
vm-test-run-inferno-ml-server-test> kill vlan (pid 7)
vm-test-run-inferno-ml-server-test> (finished: cleanup, in 0.00 seconds)
```

ngua (approved):

PR #172, opened 2025-04-21
[inferno-ml] Store model version size

ngua:
Since the `contents` of a `ModelVersion` are immutable, we can calculate its size once and store it in the DB. Currently we only need the size once in `inferno-ml-server` (when caching a model version) but now we want to display them elsewhere. Calculating it each time is wasteful.

This also includes a few other changes:
- `SqlError`s are now caught and rethrown so they show up as error traces and in error responses 
- The `unique` constraint on `models` has been changed so that a new model can be created with the same name as (a) previously terminated model(s)

### DIFF
--- a/nix/inferno-ml/migrations/v1-create-tables.sql
+++ b/nix/inferno-ml/migrations/v1-create-tables.sql
@@ -26,7 +26,7 @@ create table if not exists models
   , updated timestamptz
     -- See note above
   , terminated timestamptz
-  , unique (name, gid)
+  , unique nulls not distinct (name, gid, terminated)
   );
 
 create table if not exists mversions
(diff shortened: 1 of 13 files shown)


ngua:
NOTE: I have these changes integrated and working elsewhere; however, we shouldn't merge this quite yet until that is ready. It requires another DB wipe sadly

nix/inferno-ml/migrations/v1-create-tables.sql
```diff
@@ -42,6 +42,10 @@ create table if not exists mversions
     -- that saving a model version requires as its first step `lo_import`ing
     -- the contents
   , contents oid not null
+    -- The size of the `contents` above in bytes. This could be calculated
+    -- on-demand, but since the `contents` are immutable, we can calculate
+    -- this once and store it here
+  , size bigint not null
```
ngua, line 48: Unfortunately an `integer` column is probably not going to be big enough, as it could only represent ~2Gb in size and we might have things bigger than that

shulhi (approved):

PR #173, opened 2025-04-23
[inferno-ml] Allow non-double `Writes` and more value types

ngua:
- Makes `Writes` stream take any `IValue` (previously it was restricted to doubles only)
- Expands `IValue` to take more Inferno `Value` types (required for full range of possible writes)
- Changes Aeson instances accordingly (not as compact as before, but that's unavoidable)
- Adds `unsqueeze` primitive (tiny change so included it here)


ngua:
**NOTE**: Don't merge this yet, it needs to be integrated elsewhere

albertov (approved):

PR #174, opened 2025-05-15
[inferno-ml] Add logging primitives

ngua:
This adds the ability to log values to a "console" directly from scripts. This is more useful for Inferno ML because debugging can be difficult with the different types involved.

- Adds `Print` module to Inferno ML prelude
- Adds implementation for `inferno-ml-server`

NOTE: Previously I tried to get wildcard bindings working (e.g. `let _ = Print.print... in`) but it requires more out-of-scope work

### DIFF
--- a/inferno-core/CHANGELOG.md
+++ b/inferno-core/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Revision History for inferno-core
 *Note*: we use https://pvp.haskell.org/ (MAJOR.MAJOR.MINOR.PATCH)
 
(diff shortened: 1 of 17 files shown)


albertov:
Would a traceShow-like function avoid needing the changes needed to work-around not having statements?

ngua:
Having wildcards isn't going to work at this point. Needs more work that's not really worth the effort of avoiding unused variable warnings from Inferno LSP.

albertov (approved):

PR #218, opened 2026-05-14
[inferno-core][parser] Parser fixes

ngua:
This rewrites the parser to avoid pervasive laziness, recomputing the same information on literally every pass (I mean actually recreating the entire expression parser on every parse), and avoiding intermediate allocations (e.g. accumulating directly into strict `Map`s, rather than `[(a, b)]` then into a lazy `Map` when parsing is done). Also added some helper records and types to help clarify intent of parser code, which I honestly found difficult to follow at times.

For more specific information on decisions made, see #213, #214, #215, #216, and #217. 

All tests in `inferno` pass. The new parsers are semantically equivalent to the old ones.

At any rate, the changes have certainly paid off, as the benchmarks demonstrate. The `deep/*` benchmarks in particular show a huge improvement (this exercise longer parser chains than `wide/*`, so this makes sense). I won't paste the raw stats but a summary:

#### Wall-clock improvements

| Benchmark | Old   | New   | Speedup  |
| --------- | ----- | ----- | -------- |
| wide/100  | 304ms | 181ms | **1.7x** |
| wide/1000 | 3.93s | 2.88s | **1.4x** |
| deep/25   | 1.44s | 549ms | **2.6x** |
| deep/50   | 2.90s | 1.13s | **2.6x** |
| deep/100  | 6.08s | 2.43s | **2.5x** |

##### `wide/*` benchmark suite

| Metric              | Old      | New     | Change   |
| ------------------- | -------- | ------- | -------- |
| Total elapsed       | 85.1s    | 50.9s   | -40%     |
| Mutator time        | 38.1s    | 32.2s   | -16%     |
| GC time             | 47.0s    | 18.6s   | **-60%** |
| Productivity        | 44.9%    | 63.4%   | +18pp    |
| Heap allocated      | 314 GB   | 251 GB  | -20%     |
| Bytes copied (GC)   | 88.7 GB  | 37.2 GB | **-58%** |
| Max residency       | 501 MB   | 246 MB  | **-51%** |
| Total memory in use | 1210 MiB | 629 MiB | -48%     |
| Gen 1 max pause     | 388ms    | 173ms   | -55%     |

##### `deep/*` benchmark suite

| Metric              | Old     | New     | Change   |
| ------------------- | ------- | ------- | -------- |
| Total elapsed       | 167.2s  | 66.0s   | -60%     |
| Mutator time        | 78.0s   | 47.4s   | **-39%** |
| GC time             | 89.2s   | 18.6s   | **-79%** |
| Productivity        | 46.7%   | 71.9%   | +25pp    |
| Heap allocated      | 710 GB  | 403 GB  | -43%     |
| Bytes copied (GC)   | 177 GB  | 38.7 GB | **-78%** |
| Max residency       | 310 MB  | 78 MB   | **-75%** |
| Total memory in use | 750 MiB | 203 MiB | -73%     |
| Gen 1 max pause     | 206ms   | 56ms    | -73%     |

#### Specific changes from the old parser

- **The old parser spent more time in GC than doing work** (53% GC on deep inputs, 55% on wide). The new parser is **72% productive**.
- The old `makeExprParser` **was rebuilt on every `expr` invocation**
  - This means that _every single subexpression_ rebuilt the **ENTIRE** expression parser
  - The parser is now pre-computed once per environment and read from a lazy field (i.e. knot-tying approach)
- The old lazy `WriterT` accumulated thunk chains for comments; this has been replaced with a strict `StateT` (now GC can reclaim immediately)
- In the old apprach, `foldl` built `App` thunks in function application; this has been replaced with `foldl'`.
- the `variable` parser performed O(m) linear keyword scan _per identifier_; now O(log m) via `Set.member` on precomputed reserved keywords
- Collections (arrays, records) double-parsed their last element via `try` backtracking; this has been restructured to parse-then-decide
- Operator lists were extracted and filtered on every prefix-op parse; this is now also pre-computed in the environment

### DIFF
--- a/inferno-core/src/Inferno/Parse.hs
+++ b/inferno-core/src/Inferno/Parse.hs
@@ -3,9 +3,9 @@
 {-# LANGUAGE ExplicitForAll #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE NoFieldSelectors #-}
-
 {-# OPTIONS_GHC -Wno-unused-top-binds #-}
 
 module Inferno.Parse
@@ -27,13 +27,18 @@ where
 
 import Control.Applicative (asum, (<|>))
 import Control.Arrow ((&&&))
-import Control.Monad (void)
-import Control.Monad.Combinators.Expr (Operator, makeExprParser)
+import Control.Monad (join, void)
+import Control.Monad.Combinators.Expr
+  ( Operator (InfixL, InfixN, InfixR, Prefix),
+    makeExprParser,
+  )
 import Control.Monad.Reader (ReaderT, asks)
 import Control.Monad.State.Strict (StateT, modify')
+import Data.Bifunctor (first)
 import Data.Char (isAlphaNum, isSpace)
 import Data.Data (Data)
-import Data.Functor (($>))
+import Data.Foldable (foldl') -- foldl': DO NOT REMOVE; needed for older GHC compat
+import Data.Functor (($>), (<&>))
 import qualified Data.IntMap.Strict as IntMap
 import Data.List.NonEmpty (NonEmpty)
 import Data.Map.Strict (Map)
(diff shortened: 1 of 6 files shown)


inferno-core/src/Inferno/Parse.hs
```diff
@@ -98,516 +153,854 @@ import Text.Megaparsec
...
 
 implVarE :: Parser (Expr () SourcePos)
-implVarE = do
-  startPos <- getSourcePos
-  lexeme $ Var startPos () LocalScope . Impl . ExtIdent . Right <$> implicitVariable
+implVarE = lexeme . withSourcePos $ \pos ->
+  Var pos () LocalScope . Impl . ExtIdent . Right <$> implicitVariable
 
--- | 'intE' and 'doubleE' parse unsigned numbers
+-- | Parses unsigned integer and double literals.
```
Daniel-Diaz: I think for haddock it works better to keep the type signatures separate.

ngua: Agreed. I think this was a historical leftover (combining the signatures)

Daniel-Diaz (approved):
LGTM!

